PR #16, opened 2023-01-04
Apply tailwind

dario-piotrowicz:
> This is a preliminary step for #13

@jculvey with tailwind things do look quite similar to our current version but we do need to tweak some stuff (especially spacing), but I feel like this is good enough to get merged as is (and we can gradually improve things), please check it out locally and let me know what you think 🙂


frontend/src/theme.scss
```diff
@@ -0,0 +1,15 @@
+
```
dario-piotrowicz, line 1: @jculvey I think that we might want to remove this file as well, right?

(if so, do you prefer me to do it now or in a followup PR?)

frontend/src/styles.scss
```diff
@@ -1,5 +1,9 @@
 @use './theme.scss' as theme;
```
dario-piotrowicz, line 1: @jculvey I've renamed this file from `global` to `styles` as it looked simpler with the new folder structure (without a `styles` folder), let me know if you prefer `global` or anything else, I'm fine with whatever name you'd prefer 🙂

frontend/src/theme.scss
jculvey: Yeah, if we're not using any of that stuff anymore go for it. Better to do it now I think to make sure we've covered our bases.

frontend/src/styles.scss
jculvey: That works. The `global` bit is just a hint that these are styles that are applied globally `* { ... }, body: { ... }` etc. Now that we're using tailwind 100% there's only this file at the top level so the distinction isn't really required anymore

frontend/src/theme.scss
dario-piotrowicz: no, it is being used, that's why I kept it

so I can either remove it and remove the scss code using its sass variables, or I can just remove it and replace usages of its sass variables with the hardcoded values

I think it's mostly used for the colors, I guess I can try to completely remove it alongside the variables usages and check that it doesn't break the ui too much, if it does I can hardcode the values or try to apply some tailwind classes

dario-piotrowicz: I did a bit of both (removed things that didn't require a complex refactoring and replaced variables with hardcoded values everywhere else)

jculvey (approved):

jculvey (approved):

PR #29, opened 2023-01-06
optimize loading remote statuses

xtuc:


dario-piotrowicz (approved):
nice improvement 🙂 

but why do we need to disable favourites and reblogs?

xtuc:
@dario-piotrowicz  it's too slow.

PR #32, opened 2023-01-06
apply no-console eslint rule for BE

dario-piotrowicz:


backend/src/activitypub/activities/handle.ts
```diff
@@ -92,7 +92,6 @@ export async function handle(domain: string, activity: Activity, db: D1Database,
 	const getObjectAsId = makeGetObjectAsId(activity)
 	const getActorAsId = makeGetActorAsId(activity)
 
-	console.log(activity)
```
xtuc, line 95: this is useful for debugging, we should keep it.

xtuc (changes requested):

backend/src/activitypub/activities/handle.ts
dario-piotrowicz: I've put it back with a TODO comment for it's removal, I hope that's fine? 🙂

xtuc:
I think this is going to be annoying when debugging and I don't see what's wrong with leaving logs in prod.

dario-piotrowicz:
> I think this is going to be annoying when debugging and I don't see what's wrong with leaving logs in prod.

If you want to keep the console.logs in prod that's fine we can close the PR 🙂 

I just thought/think that they create noise (for example when running the be tests I see lots of logs which make reading the results of the tests cumbersome) and imagined that most logs are just used during development to check things but aren't needed in prod. But I am not that familiar with the BE of this project (nor BE in general) so I may be wrong if you say it's better to keep them I'm ok with it.

PR #37, opened 2023-01-08
fix: normalise uppercase github usernames

### LINKED ISSUE #36: Unable to deploy when GitHub username has uppercase characters
(Sorry for submitting this on a WIP project, we just like to play with things early 😅)

If you run through the deployment steps and have a GitHub username with any uppercase letters, it fails to deploy. This seems to be because Pages doesn't support uppercase chars in project names (and normalizes to all lowercase), and the API is case-sensitive too so fetching for `wildebeest-Cherry` produces no results even if `wildebeest-cherry` exists.

The terraform step fails with:
```
╷
│ Error: error creating cloudflare pages project "wildebeest-Cherry": You have entered an invalid project name. Project names can be 1-58 lowercase characters with dashes, but cannot start/end with a dash (/). (8000003)
│ 
│   with cloudflare_pages_project.wildebeest_pages_project,
│   on main.tf line 57, in resource "cloudflare_pages_project" "wildebeest_pages_project":
│   57: resource "cloudflare_pages_project" "wildebeest_pages_project" {
│ 
╵
```
but doesn't actual exit with a non-zero exit code so continues to try and deploy in the next step. But then fails with this on the deploy step:
```
✘ [ERROR] A request to the Cloudflare API (/accounts/***/pages/projects/wildebeest-Cherry) failed.

  Project not found. The specified project name does not match any of your existing projects. [code: 8000007]
```

Cherry:
This normalises the GitHub username in both the Action and Terraform config to lowercase. Terraform nicely supports `lower()` but GitHub Actions don't have a nice way to do this. I used the solution proposed at https://github.com/orgs/community/discussions/27086#discussioncomment-3254548 but would welcome any feedback.

Resolves #36

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -24,6 +24,14 @@ jobs:
           CF_ZONE_ID: ${{ secrets.CF_ZONE_ID }}
           CF_DEPLOY_DOMAIN: ${{ secrets.CF_DEPLOY_DOMAIN }}
 
+      # this is needed to get the lowercase version of the actor name
+      # TODO: switch to some lowercase function in the future when Actions supports it
+      - name: Set lowercase actor name
+        run: |
+          echo "ACTOR_LOWER=${GH_ACTOR,,}" >>${GITHUB_ENV}
+        env:
+          GH_ACTOR: ${{ github.actor }}
+
       - uses: actions/checkout@v2
       - uses: hashicorp/setup-terraform@v2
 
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy.yml
```diff
@@ -24,6 +24,14 @@ jobs:
           CF_ZONE_ID: ${{ secrets.CF_ZONE_ID }}
           CF_DEPLOY_DOMAIN: ${{ secrets.CF_DEPLOY_DOMAIN }}
 
+      # this is needed to get the lowercase version of the actor name
+      # TODO: switch to some lowercase function in the future when Actions supports it
+      - name: Set lowercase actor name
+        run: |
+          echo "ACTOR_LOWER=${GH_ACTOR,,}" >> ${GITHUB_ENV}
```
xtuc, line 31: The change makes sense to me. However this bash syntax is terriblee, we could use this instead: `| awk '{ print tolower($0) }'`

PR #45, opened 2023-01-09
terraform: explicitly disable auto minify of JS

elithrar:
Auto-minify breaks the Mastodon front-end. This explicitly sets the zone config to disable it.

Ref: https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs/resources/zone_settings_override

### DIFF
--- a/tf/main.tf
+++ b/tf/main.tf
@@ -135,7 +135,7 @@ resource "cloudflare_ruleset" "wildebeest_config_rules" {
         "js": false
       }
     }
-    expression  = "(http.host eq \"${cloudflare_deploy_domain}\")"
+    expression  = "(http.host eq \"${trimspace(var.cloudflare_deploy_domain)}\")"
     description = "Disable JS minification for the Mastodon subdomain"
     enabled     = true
   }
--- a/tf/main.tf
+++ b/tf/main.tf
@@ -136,7 +136,7 @@ resource "cloudflare_ruleset" "wildebeest_config_rules" {
       }
     }
     expression  = "(http.host eq \"${trimspace(var.cloudflare_deploy_domain)}\")"
-    description = "Disable JS minification for the Mastodon subdomain"
+    description = "Disable JS minification for the Wildebeest subdomain"
     enabled     = true
   }
 }
--- a/tf/main.tf
+++ b/tf/main.tf
@@ -135,7 +135,7 @@ resource "cloudflare_ruleset" "wildebeest_config_rules" {
         "js": false
       }
     }
-    expression  = "(http.host eq \"${cloudflare_deploy_domain}\""
+    expression  = "(http.host eq \"${cloudflare_deploy_domain}\")"
     description = "Disable JS minification for the Mastodon subdomain"
     enabled     = true
   }
--- a/tf/main.tf
+++ b/tf/main.tf
@@ -122,14 +122,21 @@ resource "cloudflare_access_policy" "policy" {
   }
 }
 
-resource "cloudflare_zone_settings_override" "wildebeest_zone_config" {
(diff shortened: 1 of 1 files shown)


tf/main.tf
```diff
@@ -121,3 +121,15 @@ resource "cloudflare_access_policy" "policy" {
...
 }
+
+resource "cloudflare_zone_settings_override" "wildebeest_zone_config" {
+    zone_id = var.cloudflare_zone_id
+    settings {
+        brotli = "on"
+        minify {
+            css = "on"
+            js = "off"
+            html = "off"
```
xtuc, line 132: HTML minification shouldn't break mastodon/wildebeest, it's safer than CSS/JS.

tf/main.tf
```diff
@@ -121,3 +121,15 @@ resource "cloudflare_access_policy" "policy" {
     email = ["CHANGEME@example.com"]
   }
 }
+
+resource "cloudflare_zone_settings_override" "wildebeest_zone_config" {
+    zone_id = var.cloudflare_zone_id
+    settings {
+        brotli = "on"
+        minify {
+            css = "on"
```
xtuc, line 130: why brotli/css is on? Can't we just switch JS off?

tf/main.tf
elithrar: "Shouldn't" or "won't" — I agree it's safer but have we tested?

tf/main.tf
elithrar: I'd prefer to run with Brotli on and keep HTML unminified (better for the web; less issues). We should be enabling Brotli by default anyway.

xmflsct:
Shall we apply these overrides using Configuration Rules? So to keep any existing settings untouched.

xtuc:
yes @xmflsct . One of my concern is touching the zone's setting might not be expected by the user.

xtuc (changes requested):
We should use a page rule to avoid changing someone zone settings

petebacondarwin (commented):
I don't know enough about Terraform and the cloudflare plugins to approve this, but it looks great to me! Thanks.

tf/main.tf
```diff
@@ -121,3 +121,22 @@ resource "cloudflare_access_policy" "policy" {
...
+
+  rules {
+    action = "set_config"
+    action_parameters {
+      "autominify": {
+        "js": false
+      }
+    }
+    expression  = "(http.host eq \"${trimspace(var.cloudflare_deploy_domain)}\")"
+    description = "Disable JS minification for the Mastodon subdomain"
```
xtuc, line 139:
```suggestion
    description = "Disable JS minification for the Wildebeest subdomain"
```

xtuc:
We also need to update the documentation for generating the API token, new permissions are needed for changing zone settings.

tf/main.tf
```diff
@@ -121,3 +121,22 @@ resource "cloudflare_access_policy" "policy" {
...
+  name        = "Config rules ruleset"
+  kind        = "zone"
+  phase       = "http_config_settings"
+
+  rules {
+    action = "set_config"
+    action_parameters {
+      "autominify": {
+        "js": false
+      }
```
jacobbednarz, line 136: just need to tidy up the syntax on this one

```suggestion
      autominify {
        js = false
      }
```

jacobbednarz:
using a configuration ruleset is the correct way to achieve this. you don't want to rely on page rules given they have a limited lifespan thanks to ERE and the replacement phases. in saying that though, this may be impacted by an issue in the SDKv2 where it cannot differentiate unset/false values (which we are addressing with a move to the framework provider) for `cloudflare_ruleset`.

PR #51, opened 2023-01-09
adjust wildebeest logo

### LINKED ISSUE #46: Adjust wildebeest logo
Screenshot 2023-01-09 at 15 46 09

We should be using the wildebeest logo svg file with text included so that it more versatile and easier to resize (`Wildebeest dark full`)

we also need to make sure that the WildebeestLogo component can accept different sizes and apply the correct side to the different places in which it is applied (small for the mobile header, medium small for the right column and medium for the start-instance and first-login pages)

dario-piotrowicz:
- use the wildebeest logo which includes the text as well
- make the wildebeest logo resizable via props
- delete the no longer needed mastodon logo

resolves #46


frontend/src/components/MastodonLogo.tsx
```diff
@@ -1,96 +1,81 @@
...
 } as const
 
 type Props = {
 	size: keyof typeof heightsMap
 }
 
-// TODO: the mastodonLogo is likely no longer needed
-//       we're keeping it around just in case but before
-//       launch this should be removed
-export const MastodonLogo = component$<Props>(({ size }) => {
```
dario-piotrowicz, line 15: @jculvey since I was tweaking the heightsMap values I figured I could also just remove the Mastodon logo as I imagine we aren't going to need it at this point (and it may look off with the new height values)

jculvey (approved):

petebacondarwin (approved):

PR #56, opened 2023-01-10
Linting improvements

dario-piotrowicz:
- add `lint:functions` script and make it pass\
__(sorry I wasn't checking the functions code before)__

- move be eslint rules we want off before linting comment\
 __(there is a comment in the eslintrc file indicating the rules we only temporarily turned off, I'm moving a couple of rules above that since by recent PRs it seems like we don't want those enabled)__
 
- add `lint:frontend` script

- make backend linting pass\
 __(it seems it regressed)__

- update lint script\
 __(we don't want it to be `eslint .` because we have a different eslint configuration for the frontend (more strict))__
 
- add linting checks to PRs checks


functions/start-instance-test-access.ts
```diff
@@ -3,15 +3,14 @@ import * as access from 'wildebeest/backend/src/access'
...
 
-export const onRequestGet: PagesFunction<Env, any, ContextData> = async ({ request, data, env }) => {
-	return handleGetRequest(env.DATABASE, request)
+export const onRequestGet: PagesFunction<Env, any, ContextData> = async ({ request, env }) => {
+	return handleGetRequest(env, request)
 }
 
 // Route to test if Access has been configured properly
-export async function handleGetRequest(db: D1Database, request: Request): Promise<Response> {
+export async function handleGetRequest(env: Env, request: Request): Promise<Response> {
```
xtuc, line 12: why using Env instead of D1Database? that makes the handler harder to test

functions/ap/users/[id]/outbox.ts
```diff
@@ -17,6 +13,7 @@ const headers = {
 	'content-type': 'application/json; charset=utf-8',
 }
 
+// eslint-disable-next-line @typescript-eslint/no-unused-vars -- TODO: use userKEK
```
Comment (dario-piotrowicz) [inline functions/ap/users/[id]/outbox.ts:16]: @xtuc here and in some other places where I was unsure if the variables are needed or not I left the variables and added these TODO eslint disabling comments, let me know if that's ok (or if we can get rid of the variables)

petebacondarwin (approved):

functions/start-instance-test-access.ts
xtuc: > Why does it make it harder to test?
Because to test the handler we don't need to construct an env object, we can only pass a database.

dario-piotrowicz: > Because to test the handler we don't need to construct an env object, we can only pass a database.
I am not sure I get what's harder about that, instead of passing `d1` you'd be passing `{ d1 } as Env`, doesn't seem much harder to me 😕 

But anyways the function does use the `env` so passing it is necessary regardless, isn't it?

Or again, am I missing something and an `env` object is globally available or something like that? (that in my opinion would make the function harder to test as you'd need to set the value before and after calling the function)

xtuc: Oh I see what you mean now. This endpoint isn't used anymore, that's why.

xtuc: can you please remove it in a followup PR?

xtuc (approved):

PR #66, opened 2023-01-10
More UI tests

jculvey:
Adds another ui test for the individual toot view and fixes a few things that were affecting the lighthouse score

### DIFF
--- a/frontend/src/components/layout/RightColumn/RightColumn.tsx
+++ b/frontend/src/components/layout/RightColumn/RightColumn.tsx
@@ -23,7 +23,7 @@ export default component$(() => {
 
 		return (
 			<Link href={linkTarget} class={classList}>
-				<i class={`fa ${iconName} fa-fw md:mr-3`} />
+				<i style={{ width: '1.25rem', height: '1rem' }} class={`fa ${iconName} fa-fw md:mr-3 w-5 h-4`} />
 				<span class="hidden md:inline">{linkText}</span>
 			</Link>
 		)
--- a/frontend/src/routes/(frontend)/explore/layout.tsx
+++ b/frontend/src/routes/(frontend)/explore/layout.tsx
@@ -43,7 +43,7 @@ export default component$(() => {
 		<div class="explore-wrapper flex flex-col">
 			<StickyHeader>
 				<h2 class="text-reg text-md m-0 p-4 bg-wildebeest-700 xl:rounded-t">
-					<i class="fa fa-hashtag fa-fw mr-3" />
+					<i style={{ width: '1.25rem', height: '1rem' }} class="fa fa-hashtag fa-fw mr-3 w-5 h-4" />
 					<span>Explore</span>
 				</h2>
 			</StickyHeader>
--- a/frontend/test/posts-page.spec.ts
+++ b/frontend/test/posts-page.spec.ts
@@ -22,10 +22,10 @@ describe('Toot details', () => {
 		const exploreResponse = await fetch('http://0.0.0.0:6868/explore/')
 		const exploreBody = await exploreResponse.text()
 		const match = exploreBody.match(/href="\/(@georgetakei\/[0-9a-z-]*)"/)
-		if (!match || !match[1]) fail()
 
 		// Fetch the page for it and validate the result
-		const tootPath = match[1]
+		const tootPath = match?.[1]
+		expect(tootPath).toBeTruthy()
 		const response = await fetch(`http://0.0.0.0:6868/${tootPath}`)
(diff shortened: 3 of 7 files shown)


frontend/src/components/layout/RightColumn/RightColumn.tsx
```diff
@@ -23,7 +23,7 @@ export default component$(() => {
 
 		return (
 			<Link href={linkTarget} class={classList}>
-				<i class={`fa ${iconName} fa-fw md:mr-3`} />
+				<i style={{ width: '20px', height: '16px' }} class={`fa ${iconName} fa-fw md:mr-3`} />
```
dario-piotrowicz, line 26:
```suggestion
				<i class={`fa ${iconName} fa-fw md:mr-3 w-5 h-4`} />
```

frontend/src/routes/(frontend)/explore/layout.tsx
```diff
@@ -43,7 +43,7 @@ export default component$(() => {
 		<div class="explore-wrapper flex flex-col">
 			<StickyHeader>
 				<h2 class="text-reg text-md m-0 p-4 bg-wildebeest-700 xl:rounded-t">
-					<i class="fa fa-hashtag fa-fw mr-3" />
+					<i style={{ width: '20px', height: '16px' }} class="fa fa-hashtag fa-fw mr-3" />
```
dario-piotrowicz, line 46:
```suggestion
					<i class="fa fa-hashtag fa-fw mr-3 w-5 h-4" />
```

xtuc (approved):

frontend/test/posts-page.spec.ts
```diff
@@ -15,3 +15,20 @@ describe('Posts timeline page', () => {
...
+describe('Toot details', () => {
+	it('should show an individual toot', async () => {
+		// Find a specific toot in the list
+		const exploreResponse = await fetch('http://0.0.0.0:6868/explore/')
+		const exploreBody = await exploreResponse.text()
+		const match = exploreBody.match(/href="\/(@georgetakei\/[0-9a-z-]*)"/)
+		if (!match || !match[1]) fail()
+
+		// Fetch the page for it and validate the result
+		const tootPath = match[1]
```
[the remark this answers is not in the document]
jculvey: Happy to change it to an expect.

`> generate more generic/difficult to diagnose` I actually don't think thats the case. With the proposed changes, the error you get is this:
image

as opposed to something failing on the exact line where the issues actually is (here, it's the call to `.match`). I don't think `fail` is better, they're just both equally opaque.

frontend/src/components/layout/RightColumn/RightColumn.tsx
jculvey: Will this still resolve the CLS issue we're trying to fix? I thought the recommended practice was to set explicit width/height on elements as opposed to waiting for a (potentially bulky css file like tailwind or font-awesome) to download since this will cause a reflow.

https://web.dev/optimize-cls/#images-without-dimensions

frontend/test/posts-page.spec.ts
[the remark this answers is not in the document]
dario-piotrowicz: I am not sure, I guess it can be a personal preference thing

for me it is clear that you are looking for an element and you make sure/expect that it exists

if you prefer the fail I won't insist 🙂 

By the way, I just tried to use `fail` and it is failing the test because the `fail` method is not defined (not because the method makes the test fail), does that happen on your end as well?
Screenshot 2023-01-10 at 20 36 14

frontend/src/components/layout/RightColumn/RightColumn.tsx
dario-piotrowicz: ah I see, thanks a bunch for the article 🙂 

it says that the best practice is to set the `width` and `height` attributes (not setting them with styles directly), no?

I think css kicks in afterwards, so I think that the best thing would be to do something like:
```suggestion
				<i width="20" height="16" class={`fa ${iconName} fa-fw md:mr-3 w-5 h-4`} />
```
so this would set the initial width and height for the image and then would override it with w-5 and h-4 when the css is ready (w-5 and h-4 are respectively 20px and 16px for the medium font-size so we won't see any difference in that case, but then they will scale properly based on the browser's font size)

alternatively if you prefer you can set directly the sizes in rem directly:
```suggestion
				<i style={{ width: '5rem', height: '4rem' }} class={`fa ${iconName} fa-fw md:mr-3 w-5 h-4`} />
```

dario-piotrowicz (approved):

PR #67, opened 2023-01-10
improve styling

dario-piotrowicz:
improve styling by:
 - removing unnecessary css (rely on tailwind instead)
 - remove usage of pixels (use rem instead)
 - update state- colors to wildebeest- colors (regression)


frontend/src/components/Status/index.scss
```diff
@@ -1,3 +1,7 @@
+// Important: The rules present in this file apply to dynamic content
```
dario-piotrowicz, line 1: @jculvey I think it would be really cool if instead of using css here for the content, if we parsed the html and replaced and applied tailwind classes, but I think that this wouldn't be trivial, so it's probably not worth it going down that rabbit hole 😅

xtuc (approved):

PR #70, opened 2023-01-11
Use Pleroma-compatible version string

mkljczk:
As documented here: https://gitlab.com/soapbox-pub/soapbox/-/blob/develop/docs/development/developing-backend.md

This is implemented by other non-Mastodon backends utilizing Mastodon API, including Pixelfed, Pleroma, Mitra, Friendica…


xtuc:
Thanks @mkljczk. Does Mastodon-itself uses the version?

xtuc:
Closing in favor of https://github.com/cloudflare/wildebeest/pull/119

PR #71, opened 2023-01-11
remove before, after css code

dario-piotrowicz:
I am slightly on the fence on this one, I wonder if the css is just simpler here, or if, once using tailwind it is considered better to avoid css as much as possible 🤔

@jculvey what's your take on it?

> PS: I googled it and I could not find a definitive answer on whether using css with tailwind is a bad practice


frontend/src/routes/(frontend)/explore/layout.tsx
```diff
@@ -54,3 +52,35 @@ export default component$(() => {
...
+	'before:bottom-[-1px]',
+	'before:left-1/2',
+	'before:translate-x-[-50%]',
+	'before:border-solid',
+	'before:border-t-0',
+	'before:border-x-[0.7rem]',
+	'before:border-b-[0.7rem]',
+	'before:border-x-transparent',
+	'before:border-b-wildebeest-500',
+	'after:block',
```
dario-piotrowicz, line 72: I would have liked to define the repeating classes only once and combined generated them dynamically, unfortunately is seems like you need to have the full class present for tailwind to pick it up 😓  (/you can construct the class names dynamically)

https://tailwindcss.com/docs/content-configuration#dynamic-class-names

jculvey:
I think it's ok to use actual CSS classes here and there for the complicated use cases as opposed to having a tailwind only rule. It does seem a bit simpler here.

I think it's perhaps possible to have the best of both worlds though. The `theme` function exported by tailwind is globally available in our scss files (I think we were talking about this earlier), so you could just keep the scss classes and redefine their rules using tailwind theme values instead of hardcoded ones. https://tailwindcss.com/docs/using-with-preprocessors

jculvey (approved):
If it's working then no need to change it, I think both are viable approaches

PR #72, opened 2023-01-11
MOW-95: improve one-click deployement

xtuc:


scripts/generate-one-click-deploy-button.mjs
```diff
@@ -0,0 +1,35 @@
...
+		name: 'Instance description',
+		secret: 'INSTANCE_DESCR',
+		descr: 'A short, plain-text description of your instance',
+	},
+]
+
+const fields = FIELDS.map((x) => JSON.stringify(x))
+	.map((v) => `fields=${v}`)
+	.join('&')
+const url = new URL(`/?url=${PROJECT_URL}&authed=true&${fields}`, ONE_CLICK_BASE_URL)
```
dario-piotrowicz, line 34: since the fields to in the url and there are spaces in the descriptions, don't we need to encode the field?(with [encodeURIComponent](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent) for example)

dario-piotrowicz (approved):
nice! 🤩

scripts/generate-one-click-deploy-button.mjs
xtuc: it gets encoded

dario-piotrowicz: ah ok cool thanks 🙂

PR #75, opened 2023-01-11
improve be (+functions) linting:

dario-piotrowicz:
- enable no-unsafe-call for be/functions linting
 - enable backend await-thenable eslint rule
 - enable backend no-misused-promises eslint rule

### DIFF
--- a/backend/src/access/index.ts
+++ b/backend/src/access/index.ts
@@ -121,7 +121,7 @@ export const generateValidator =
 
 		const unroundedSecondsSinceEpoch = Date.now() / 1000
 
-		const payloadObj = JSON.parse(textDecoder.decode(base64URLDecode(payload)))
+		const payloadObj = JSON.parse(textDecoder.decode(base64URLDecode(payload))) as JWTPayload
 
 		// For testing disable JWT checks.
 		// Ideally we match the production behavior in testing but that
--- a/backend/src/activitypub/activities/handle.ts
+++ b/backend/src/activitypub/activities/handle.ts
@@ -29,7 +29,7 @@ function extractID(domain: string, s: string | URL): string {
 	return s.toString().replace(`https://${domain}/ap/users/`, '')
 }
 
-export function makeGetObjectAsId(activity: Activity): Function {
+export function makeGetObjectAsId(activity: Activity) {
 	return () => {
 		let url: any = null
 		if (activity.object.id !== undefined) {
@@ -55,7 +55,7 @@ export function makeGetObjectAsId(activity: Activity): Function {
 	}
 }
 
-export function makeGetActorAsId(activity: Activity): Function {
+export function makeGetActorAsId(activity: Activity) {
 	return () => {
 		let url: any = null
 		if (activity.actor.id !== undefined) {
--- a/backend/src/cache/index.ts
+++ b/backend/src/cache/index.ts
@@ -1,3 +1,5 @@
+import type { Env } from 'wildebeest/consumer/src'
+
 const CACHE_DO_NAME = 'cachev1'
 
 export interface Cache {
@@ -5,7 +7,7 @@ export interface Cache {
 	put<T>(key: string, value: T): Promise<void>
 }
 
-export function cacheFromEnv(env: any): Cache {
+export function cacheFromEnv(env: Env): Cache {
 	return {
 		async get<T>(key: string): Promise<T | null> {
 			const id = env.DO_CACHE.idFromName(CACHE_DO_NAME)
--- a/backend/src/mastodon/notification.ts
+++ b/backend/src/mastodon/notification.ts
@@ -8,7 +8,11 @@ import { getPersonById } from 'wildebeest/backend/src/activitypub/actors'
 import type { WebPushInfos, WebPushMessage } from 'wildebeest/backend/src/webpush/webpushinfos'
 import { WebPushResult } from 'wildebeest/backend/src/webpush/webpushinfos'
 import type { Actor } from 'wildebeest/backend/src/activitypub/actors'
-import type { NotificationType, Notification } from 'wildebeest/backend/src/types/notification'
+import type {
+	NotificationType,
+	Notification,
+	NotificationsQueryResult,
+} from 'wildebeest/backend/src/types/notification'
 import { getSubscriptionForAllClients } from 'wildebeest/backend/src/mastodon/subscription'
 import type { Cache } from 'wildebeest/backend/src/cache'
 
@@ -24,10 +28,10 @@ export async function createNotification(
           VALUES (?, ?, ?, ?)
           RETURNING id
 `
-	const row: { id: string } = await db
+	const row = await db
 		.prepare(query)
 		.bind(type, actor.id.toString(), fromActor.id.toString(), obj.id.toString())
-		.first()
+		.first<{ id: string }>()
 	return row.id
 }
 
@@ -39,7 +43,7 @@ export async function insertFollowNotification(db: D1Database, actor: Actor, fro
           VALUES (?, ?, ?)
           RETURNING id
 `
-	const row: { id: string } = await db.prepare(query).bind(type, actor.id.toString(), fromActor.id.toString()).first()
+	const row = await db.prepare(query).bind(type, actor.id.toString(), fromActor.id.toString()).first<{ id: string }>()
 	return row.id
 }
 
@@ -187,7 +191,7 @@ export async function getNotifications(db: D1Database, actor: Actor, domain: str
   `
 
 	const stmt = db.prepare(query).bind(actor.id.toString())
-	const { results, success, error } = await stmt.all()
+	const { results, success, error } = await stmt.all<NotificationsQueryResult>()
 	if (!success) {
 		throw new Error('SQL error: ' + error)
 	}
@@ -198,7 +202,7 @@ export async function getNotifications(db: D1Database, actor: Actor, domain: str
 	}
 
 	for (let i = 0, len = results.length; i < len; i++) {
-		const result = results[i] as any
+		const result = results[i]
 		const properties = JSON.parse(result.properties)
 		const notifFromActorId = new URL(result.notif_from_actor_id)
 
--- a/backend/src/types/notification.ts
+++ b/backend/src/types/notification.ts
@@ -1,5 +1,6 @@
 import type { MastodonAccount } from 'wildebeest/backend/src/types/account'
 import type { MastodonStatus } from 'wildebeest/backend/src/types/status'
+import type { ObjectsRow } from './objects'
 
 export type NotificationType =
 	| 'mention'
@@ -19,4 +20,13 @@ export type Notification = {
 	created_at: string
 	account: MastodonAccount
 	status?: MastodonStatus
+}
+
+export interface NotificationsQueryResult extends ObjectsRow {
+	type: NotificationType
+	original_actor_id: URL
+	notif_from_actor_id: URL
+	notif_cdate: string
+	notif_id: URL
+	from_actor_id: string
 }
--- a/backend/src/utils/body.ts
+++ b/backend/src/utils/body.ts
@@ -26,10 +26,8 @@ export async function readBody<T>(request: Request): Promise<T> {
 			// The `key[]` notiation is used when sending an array of values.
 
 			const key2 = key.replace('[]', '')
-			if (out[key2] === undefined) {
-				out[key2] = []
-			}
-			out[key2].push(value)
+			const outArr: unknown[] = (out[key2] ??= [])
+			outArr.push(value)
 		} else {
 			out[key] = value
 		}
--- a/backend/src/utils/httpsigjs/parser.ts
+++ b/backend/src/utils/httpsigjs/parser.ts
@@ -1,4 +1,3 @@
-// @ts-nocheck
 // Copyright 2012 Joyent, Inc.  All rights reserved.
 
 import { HEADER, HttpSignatureError, InvalidAlgorithmError, validateAlgorithm } from './utils'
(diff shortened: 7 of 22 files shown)


petebacondarwin:
Needs a rebase

backend/src/utils/body.ts
```diff
@@ -29,7 +29,8 @@ export async function readBody<T>(request: Request): Promise<T> {
 			if (out[key2] === undefined) {
 				out[key2] = []
 			}
-			out[key2].push(value)
+			// eslint-disable-next-line @typescript-eslint/no-extra-semi
```
dario-piotrowicz, line 32: for some reason prettier insists in adding this semicolon 😕

petebacondarwin: The bots are fighting each other!

backend/test/activitypub/follow.spec.ts
```diff
@@ -24,7 +24,7 @@ describe('ActivityPub', () => {
 			globalThis.fetch = async (input: any) => {
 				if (input.url === `https://${domain}/ap/users/sven2/inbox`) {
 					assert.equal(input.method, 'POST')
-					const data = await input.json()
+					const data = await (input as Response).json()
```
petebacondarwin, line 27: `input` is actually a `Request` not `Response` 😱 

But I think the correct fix here is to give the function being defined a proper type:

```
			globalThis.fetch = async (input: RequestInfo) => {
				const request = new Request(input)
				if (request.url === `https://${domain}/ap/users/sven2/inbox`) {
					assert.equal(request.method, 'POST')
					const data = await request.json()
					receivedActivity = data
					console.log({ receivedActivity })
					return new Response('')
				}

				throw new Error('unexpected request to ' + request.url)
			}
```

backend/src/utils/body.ts
dario-piotrowicz: yeah! thanks! 🤩

backend/test/activitypub/follow.spec.ts
dario-piotrowicz: sorry! 😱 😭 

as I scanned thought the code and saw the `.json` I instinctively thought it was a response 😓
(I've checked and I've done that in a couple of places I've fixed all of them 😓)

By the way, awesome solution, I tend to forget that you can pass the request request infos in that way 😃 🤩

backend/src/activitypub/activities/handle.ts
```diff
@@ -108,12 +108,12 @@ export async function handle(
 			const objectId = getObjectAsId()
 
 			// check current object
-			const object = await objects.getObjectBy(db, 'original_object_id', objectId.toString())
```
petebacondarwin, line 111: I think we could/should avoid these casts by making the `makeGetObjectAsId()` function return a better type.

petebacondarwin: In fact simply removing the `Function` return type and leaving TS to infer the type resolves the problem.

backend/src/access/index.ts
```diff
@@ -131,7 +131,7 @@ export const generateValidator =
 			if (payloadObj.iss && payloadObj.iss !== certsURL.origin) {
 				throw new Error('JWT issuer is incorrect.')
 			}
-			if (payloadObj.aud && !payloadObj.aud.includes(aud)) {
+			if (payloadObj.aud && !(payloadObj.aud as string).includes(aud)) {
```
petebacondarwin, line 134: Instead of casting here, we should cast the construction of the payloadObj on line 124 above with `as JWTPayload`.

(Or even better reuse, somehow, the `getPayload()` function rather than re-parsing it here.)

backend/src/mastodon/notification.ts
```diff
@@ -212,7 +212,7 @@ export async function getNotifications(db: D1Database, actor: Actor, domain: str
 		const notifFromAccount = await loadExternalMastodonAccount(acct, notifFromActor)
 
 		const notif: Notification = {
-			id: result.notif_id.toString(),
+			id: (result.notif_id as string).toString(),
```
petebacondarwin, line 215: If we are casting to string, then why call `toString()` on it :-P ?

petebacondarwin: Ideally we should provide a type to the `stmt.all()` function that defines the types expected from the query.
e.g. something like (after also defining `ObjectsRow`)

```
interface NotificationsQueryResult extends ObjectsRow {
  type,
  actor_id,
  notif_from_actor_id,
  notif_cdate,
  notif_id
}
```

backend/src/utils/httpsigjs/parser.ts
```diff
@@ -241,7 +241,7 @@ export function parseRequest(request: Request, options?: Options): ParsedSignatu
 			parsed.params.headers = ['date']
 		}
 	} else {
-		parsed.params.headers = parsed.params.headers.split(' ')
+		parsed.params.headers = (parsed.params.headers as string).split(' ')
```
petebacondarwin, line 244: Rather than casting here we should create a new `headers: string[]` variable that is assigned from the `parsed.params.headers` property.
Also that property should get a proper type above, e.g.

```
  interface ParsedRequest {
    scheme: string
    params: { headers?: string }
    signingString: string
  }
	const parsed: ParsedRequest = {
		scheme: authz === request.headers.get(HEADER.SIG) ? 'Signature' : '',
		params: {},
		signingString: '',
	}
```

backend/test/mastodon/media.spec.ts
```diff
@@ -12,8 +12,11 @@ const domain = 'cloudflare.com'
...
 	describe('media', () => {
 		test('upload image creates object', async () => {
-			globalThis.fetch = async (input: any) => {
-				if (input.url.toString() === 'https://api.cloudflare.com/client/v4/accounts/testaccountid/images/v1') {
+			globalThis.fetch = async (input: unknown) => {
+				if (
+					(input as { url: URL }).url.toString() ===
+					'https://api.cloudflare.com/client/v4/accounts/testaccountid/images/v1'
+				) {
 					return new Response(
```
petebacondarwin, line 20: Similar to `follow.spec.ts`, something like:

```ts
			globalThis.fetch = async (input: RequestInfo) => {
				const request = new Request(input)
				if (request.url === 'https://api.cloudflare.com/client/v4/accounts/testaccountid/images/v1') {
```

backend/test/mastodon/notifications.spec.ts
```diff
@@ -117,7 +117,7 @@ describe('Mastodon APIs', () => {
```
petebacondarwin, line 117: Sorry to be annoying but I am not keen on these vague `any` types, even in tests.
How about:

```ts
			globalThis.fetch = async (input: RequestInfo, data?: RequestInit) => {
				assert(data)
				const headers = new Headers(data.headers)
				if (input === 'https://push.com') {
					assert(headers.get('Authorization')?.includes('WebPush'))

					const cryptoKey = headers.get('Crypto-Key')
					assert(cryptoKey)
					const cryptoKeyHeader = parseCryptoKey(cryptoKey)
					assert(cryptoKeyHeader.dh)
					assert(cryptoKeyHeader.p256ecdsa)

					// Ensure the data has a valid signature using the client public key
					assert(data.body)
					const body = data.body as ArrayBuffer | ArrayBufferView
					const sign = await crypto.subtle.sign({ name: 'ECDSA', hash: 'SHA-256' }, clientKeys.privateKey, body)
					assert(await crypto.subtle.verify({ name: 'ECDSA', hash: 'SHA-256' }, clientKeys.publicKey, sign, body))

```

backend/test/mastodon/oauth.spec.ts
```diff
@@ -126,15 +127,15 @@ describe('Mastodon APIs', () => {
 			const location = res.headers.get('location')
 			assert.equal(location, 'https://redirect.com/a')
 
-			const actor = await db.prepare('SELECT * FROM actors').first()
+			const actor = await db.prepare('SELECT * FROM actors').first<{ properties: string; email: string; id: string }>()
```
petebacondarwin, line 130: How about 

```
   .first<{ properties: string; email: string; id: string }&Actor>()
```

then we don't need the `as unknown as Actor` below.

backend/test/mastodon/statuses.spec.ts
```diff
@@ -60,9 +61,9 @@ describe('Mastodon APIs', () => {
 			assert.equal(res.status, 200)
 			assertJSON(res)
 
-			const data = await res.json<any>()
-			assert(data.uri.includes('example.com'))
-			assert(data.uri.includes(data.id))
+			const data = await res.json<MastodonStatus>()
+			assert((data.uri as unknown as string).includes('example.com'))
```
petebacondarwin, line 65: I hate this because the type says URL but clearly the actual runtime value is a string.
How about instead of a cast we fudge it with:

```
			assert((data.uri.toString()).includes('example.com'))
```

Or perhaps it is best to leave the ` as unknown as string` to remind us that the compile time and runtime types don't line up?

functions/api/v1/notifications/[id].ts
```diff
@@ -48,7 +48,7 @@ export async function handleRequest(
 	const fromAccount = await loadExternalMastodonAccount(acct, fromActor)
 
 	const out: Notification = {
-		id: row.notif_id.toString(),
+		id: (row.notif_id as object).toString(),
```
Comment (petebacondarwin) [inline functions/api/v1/notifications/[id].ts:51]: Similar issue as `mastodon/notification.ts`

backend/src/activitypub/activities/handle.ts
dario-piotrowicz: yeah totally! (sorry I wasn't digging around the code 😓 🙇 )

backend/test/mastodon/oauth.spec.ts
dario-piotrowicz: ah yeah totally 🙂

backend/test/mastodon/notifications.spec.ts
dario-piotrowicz: ah no I totally agree, but if it's ok I would leave this to a separate PR, in this one I am just trying to enable those handful of eslint rules you can see in the eslintrc change.

For the `any`s we'll need to enable this one:
https://github.com/cloudflare/wildebeest/blob/4a6bf1739b2df758784b44625beebc16b8234600/.eslintrc.cjs#L36
which unfortunately (as I saw then I gave it a quick shot some time ago) I think it's going to be the hardest rule to enable 😓

backend/test/mastodon/media.spec.ts
dario-piotrowicz: sorry I missed this one 😓 👍

backend/test/mastodon/statuses.spec.ts
dario-piotrowicz: yeah this type of thing is quite confusing/messed-up 😭

I am not sure, whatever you think is best works for me 🙂👍 

but if you want a reminder then I think I would the go for something like:
```suggestion
                         //@ts-expect-error -- TO FIX: uri should be of type URL but in the actual runtime is a string
			assert((data.uri as unknown as string).includes('example.com'))
```

backend/src/access/index.ts
dario-piotrowicz: I am not sure how to run and test this code 😓, for that I would just do the `as JWTPayload` here? would that be ok by you?

backend/src/mastodon/notification.ts
dario-piotrowicz: I just quickly temporarily added an inline type to make eslint happy, I'll try to see if a better type and be defined and shared 👍

backend/src/utils/httpsigjs/parser.ts
dario-piotrowicz: Done, please have a look and let me know if how I fixed it is ok 🙂

functions/api/v1/notifications/[id].ts
Comment (dario-piotrowicz) [reply functions/api/v1/notifications/[id].ts]: again, I've applied a temporary solution and will try to see if a better type and be defined and shared 👍

Comment (dario-piotrowicz) [reply functions/api/v1/notifications/[id].ts]: updated, please have a look 🙂

backend/src/mastodon/notification.ts
dario-piotrowicz: I've introduced the `NotificationsQueryResult` and `ObjectsRow` interfaces, please have a look 🙂

backend/src/cache/index.ts
```diff
@@ -1,11 +1,13 @@
+import { Env } from 'wildebeest/consumer/src'
```
petebacondarwin, line 1:
```suggestion
import type { Env } from 'wildebeest/consumer/src'
```

??

backend/src/types/notification.ts
```diff
@@ -1,5 +1,6 @@
 import type { MastodonAccount } from 'wildebeest/backend/src/types/account'
 import type { MastodonStatus } from 'wildebeest/backend/src/types/status'
+import { ObjectsRow } from './objects'
```
petebacondarwin, line 3:
```suggestion
import type { ObjectsRow } from './objects'
```

??

backend/src/utils/httpsigjs/parser.ts
```diff
@@ -352,14 +361,15 @@ export function parseRequest(request: Request, options?: Options): ParsedSignatu
 	headers.forEach(function (hdr) {
 		// Remember that we already checked any headers in the params
 		// were in the request, so if this passes we're good.
-		if (parsed.params.headers.indexOf(hdr.toLowerCase()) < 0)
-			throw new MissingHeaderError(hdr + ' was not a signed header')
+		if (parsedHeaders.indexOf(hdr.toLowerCase()) < 0) throw new MissingHeaderError(hdr + ' was not a signed header')
```
petebacondarwin, line 364:
```suggestion
		if (parsedHeaders.indexOf(hdr.toLowerCase()) < 0) {
		  throw new MissingHeaderError(hdr + ' was not a signed header')
		}
```

??

petebacondarwin (approved):
OK - this is a great improvement to the code typings. Let's land it.

PR #82, opened 2023-01-11
Hide tags and news routes

jculvey:
We're deferring these for the MVP so hiding the pages (that currently use hardcoded data).

Also fixes an alignment issue with the globe icon on statuses


dario-piotrowicz (approved):
I was supposed to fix the globe alignment! 😭

PR #87, opened 2023-01-12
add git config fix for build CI/CD git issue

dario-piotrowicz:


.github/workflows/deploy.yml
```diff
@@ -184,6 +184,7 @@ jobs:
           apiToken: ${{ secrets.CF_API_TOKEN }}
           preCommands: |
             echo "*** pre commands ***"
+            git config --global --add safe.directory /github/workspace
```
petebacondarwin, line 187: Is `/github/workspace` the actual path that we use or just some boilerplate that was in the error message?

petebacondarwin (approved):

.github/workflows/deploy.yml
dario-piotrowicz: > Is /github/workspace the actual path that we use or just some boilerplate that was in the error message?

ah... yeah, I'm not sure 😓

we can merge and give it a try?

or I could use `${{ github.workspace }}` instead:
[Screenshot 2023-01-12 at 11 14 54](https://docs.github.com/en/actions/learn-github-actions/contexts)
?

dario-piotrowicz: I think @xtuc suggested that it is coming from wrangler?
https://github.com/cloudflare/wrangler-action/blob/master/Dockerfile#L3

xtuc: can't we find another way to getting the git hash?

dario-piotrowicz: sure, @petebacondarwin suggested that in the chat, let me have a look...

dario-piotrowicz:
Closing this PR in favour of https://github.com/cloudflare/wildebeest/pull/93

PR #93, opened 2023-01-12
fix commit hash setting

dario-piotrowicz:


xtuc (approved):

package.json
```diff
@@ -36,7 +36,7 @@
 		"pages": "NO_D1_WARNING=true wrangler pages",
 		"database:migrate": "yarn d1 migrations apply DATABASE",
 		"database:create-mock": "rm -f .wrangler/state/d1/DATABASE.sqlite3 && yarn database:migrate --local && node ./frontend/mock-db/run.mjs",
-		"dev": "yarn build && yarn database:migrate --local && yarn pages dev frontend/dist --d1 DATABASE --persist --compatibility-date=2022-12-20",
+		"dev": "export COMMIT_HASH=$(git rev-parse HEAD) && yarn build && yarn database:migrate --local && yarn pages dev frontend/dist --d1 DATABASE --persist --compatibility-date=2022-12-20",
```
petebacondarwin, line 39: Do we really care about the hash in dev mode?
How about we just hard code something like:

```ts
const commitHash = process.env['COMMIT_HASH']?.slice(0, 7) ?? "local-dev"
```

petebacondarwin (approved):

PR #96, opened 2023-01-12
add page not found

dario-piotrowicz:
Adding a 404 page:
Screenshot 2023-01-12 at 15 37 58

I've copied mastodon 404 page but without an image:
Screenshot 2023-01-12 at 15 38 17

___

The routes all work like mastodon, the only difference with mastodon is that here I am redirecting to `/not-found` when a page is not found instead of keeping the same location but displaying the 404 page. This implementation is cleaner and simpler, but if we prefer to follow more closely mastodon I can implement it in the same way they do

### DIFF
--- a/frontend/src/routes/(frontend)/[accountId]/index.tsx
+++ b/frontend/src/routes/(frontend)/[accountId]/index.tsx
@@ -1,8 +1,22 @@
 import { component$ } from '@builder.io/qwik'
-import { useLocation } from '@builder.io/qwik-city'
+import { loader$ } from '@builder.io/qwik-city'
+
+export const accountLoader = loader$(({ redirect, request }) => {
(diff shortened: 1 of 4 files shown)


frontend/src/routes/(frontend)/[accountId]/index.tsx
```diff
@@ -1,8 +1,22 @@
...
+})
 
 export default component$(() => {
-	const location = useLocation()
+	const accountDetails = accountLoader.use()
 
-	return <div>account details {location.params.accountId}</div>
+	// TODO: Implement the account view
+	return <>{accountDetails.value && <div>account details</div>}</>
 })
```
Comment (xtuc) [inline frontend/src/routes/(frontend)/[accountId]/index.tsx:22]: we could just delete  this route

xtuc (approved):

PR #109, opened 2023-01-16
ignore duplicated follow

xtuc:


backend/src/mastodon/follow.ts
```diff
@@ -9,7 +9,7 @@ export async function addFollowing(db: D1Database, actor: Actor, target: Actor,
 	const id = crypto.randomUUID()
 
 	const query = `
-		INSERT INTO actor_following (id, actor_id, target_actor_id, state, target_actor_acct)
+		INSERT OR IGNORE INTO actor_following (id, actor_id, target_actor_id, state, target_actor_acct)
```
dario-piotrowicz, line 12: instead alongside only ignoring the request, could we also return an error/warning to the user?

dario-piotrowicz (approved):

PR #115, opened 2023-01-16
improve the frontend lint coverage

dario-piotrowicz:
by including in the lint check not only src
but also mock-db, test and adaptors

### DIFF
--- a/frontend/mock-db/init.ts
+++ b/frontend/mock-db/init.ts
@@ -5,13 +5,16 @@ import { statuses } from 'wildebeest/frontend/src/dummyData'
 import type { Account, MastodonStatus } from 'wildebeest/frontend/src/types'
 
 const kek = 'test-kek'
+/* eslint-disable @typescript-eslint/no-empty-function */
 const queue = {
 	async send() {},
 	async sendBatch() {},
 }
-const kv_cache: any = {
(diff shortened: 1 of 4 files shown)


frontend/mock-db/init.ts
```diff
@@ -4,14 +4,27 @@ import * as reblogAPI from 'wildebeest/functions/api/v1/statuses/[id]/reblog'
...
 import type { Account, MastodonStatus } from 'wildebeest/frontend/src/types'
 
+/* eslint-disable @typescript-eslint/no-empty-function */
 const kek = 'test-kek'
 const queue = {
 	async send() {},
 	async sendBatch() {},
 }
-const kv_cache: any = {
+const kv_cache: KVNamespace = {
```
xtuc, line 13: Can we leave the any here? The other KV methods will never be called, it's dead code.

dario-piotrowicz: yeah... I did feel like it was a bit too much...

ok I can revert it back 😢

dario-piotrowicz: actually, I removed the `: any` and added a cast when we're using the variable, I hope that's fine 🙂

xtuc (approved):

PR #118, opened 2023-01-17
MOW-102 - Add sanitization to ActivityPub Objects

petebacondarwin:


backend/src/activitypub/objects/index.ts
```diff
@@ -29,33 +29,33 @@ export function uri(domain: string, id: string): URL {
...
 	db: D1Database,
 	type: string,
 	properties: any,
 	originalActorId: URL,
 	local: boolean
-): Promise<Object> {
+): Promise<Type> {
 	const uuid = crypto.randomUUID()
 	const apId = uri(domain, uuid).toString()
+	const sanitizedObject = await sanitizeObjectProperties(properties)
```
dario-piotrowicz, line 42:
```suggestion
	const sanitizedProperties = await sanitizeObjectProperties(properties)
```

backend/src/activitypub/objects/index.ts
```diff
@@ -78,11 +78,13 @@ type CacheObjectRes = {
 export async function cacheObject(
 	domain: string,
 	db: D1Database,
-	properties: any,
+	properties: unknown,
 	originalActorId: URL,
 	originalObjectId: URL,
 	local: boolean
 ): Promise<CacheObjectRes> {
+	const sanitizedObject = await sanitizeObjectProperties(properties)
```
dario-piotrowicz, line 86:
```suggestion
	const sanitizedProperties = await sanitizeObjectProperties(properties)
```

backend/src/activitypub/objects/index.ts
```diff
@@ -179,3 +181,71 @@ WHERE objects.${key}=?
 		originalObjectId: result.original_object_id,
 	} as Object
 }
+
+/** Is the given `value` an ActivityPub Object? */
```
dario-piotrowicz, line 185: is this comment beneficial? (also in the function there isn't anything ActivityPub specific, so I feel like the comment is a bit off?)

backend/src/activitypub/objects/index.ts
```diff
@@ -179,3 +181,71 @@ WHERE objects.${key}=?
...
+	return await contentRewriter.transform(new Response(unsafeContent)).text()
+}
+
+/**
+ * Sanitizes given string as an ActivityPub Object name.
+ *
+ * This sanitization removes all HTML elements from the string leaving only the text content.
+ */
+export async function sanitizeName(dirty: string): Promise<string> {
+	return await nameRewriter.transform(new Response(dirty)).text()
```
dario-piotrowicz, line 226:
```suggestion
export async function sanitizeName(unsafeName: string): Promise<string> {
	return await nameRewriter.transform(new Response(unsafeName)).text()
```
(for consistency with the `sanitizeContent` function)

backend/src/activitypub/objects/index.ts
```diff
@@ -179,3 +181,71 @@ WHERE objects.${key}=?
...
+contentRewriter.on('*', {
+	element(el) {
+		if (!['p', 'span', 'br', 'a'].includes(el.tagName)) {
+			el.tagName = 'p'
+		}
+
+		if (el.hasAttribute('class')) {
+			const classes = el.getAttribute('class')!.split(' ')
+			const sanitizedClasses = classes.filter((c) =>
+				/^(h|p|u|dt|e)-|^mention$|^hashtag$|^ellipsis$|^invisible$/.test(c)
```
dario-piotrowicz, line 239: `^(h|p|u|dt|e)-` will test positive with classes like `h-super-long-class-name` is that ok?

dario-piotrowicz (approved):

xtuc (approved):
Nice!

backend/src/activitypub/objects/index.ts
petebacondarwin: This file is all about ActivityPub Objects.
I appreciate that the implementation is a bit thin. I originally had more checks in place but it appears that it is possible to create AP Objects with almost no properties!
The description is accurate for its usage. It should only be used when checking whether a value is an AP Object, in order to narrow the type.
Removing the description might confuse people into using this function in non-supported scenarios, which would then break if we change or tighten up the implementation.

backend/src/activitypub/objects/index.ts
```diff
@@ -179,3 +181,71 @@ WHERE objects.${key}=?
...
+ * Sanitizes given string as an ActivityPub Object name.
+ *
+ * This sanitization removes all HTML elements from the string leaving only the text content.
+ */
+export async function sanitizeName(dirty: string): Promise<string> {
+	return await nameRewriter.transform(new Response(dirty)).text()
+}
+
+const contentRewriter = new HTMLRewriter()
+contentRewriter.on('*', {
```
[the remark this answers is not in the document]
petebacondarwin: Actually, even if the input was very large, we have no choice but to parse all the tags, since we only know the list of safe tags that should be ignored, we must check every tag.
As it turns out, the way that HTMLRewriter works, this is pretty optimal. It is not the same algorithm as DOM query selectors.

backend/src/activitypub/objects/index.ts
dario-piotrowicz: ok thanks for the clarification, I'm fine with the comment 🙂

backend/src/activitypub/objects/index.ts
petebacondarwin: According to the mastodon docs, that is what they do.

> The sanitizer will keep classes if they begin with microformats prefixes or are semantic classes

PR #119, opened 2023-01-17
Pleroma-compatible version string

xtuc:
Refs https://github.com/cloudflare/wildebeest/pull/70


celso (approved):

mkljczk:
Not really useful with commit hashes. For example, in Soapbox we parse the semver-compatible version to specify whether a given feature is provided by the backend. https://gitlab.com/soapbox-pub/soapbox/-/blob/develop/app/soapbox/utils/features.ts

xtuc:
@mkljczk thanks for the input. We are going to set a constant version (0.0.1), we don't plan to version the project before the official release.

PR #120, opened 2023-01-17
set compatibility date

xtuc:


celso (approved):

tf/main.tf
```diff
@@ -123,6 +123,8 @@ resource "cloudflare_pages_project" "wildebeest_pages_project" {
       d1_databases = {
         DATABASE = sensitive(var.d1_id)
       }
+
+      compatibility_date = "2023-01-09"
```
petebacondarwin, line 127: Why 9 Jan 2023 by the way?

petebacondarwin (approved):

tf/main.tf
xtuc: It's the one set on social.eng.chat

PR #132, opened 2023-01-18
fetch status context instead of providing an empty one

dario-piotrowicz:
As a result we can now see a toot's replies:
Screenshot 2023-01-18 at 13 45 07


frontend/src/routes/(frontend)/[accountId]/[statusId]/index.tsx
```diff
@@ -4,20 +4,25 @@ import Status from '~/components/Status'
...
+	const domain = new URL(request.url).hostname
+	const statusResponse = await statusAPI.handleRequest(platform.DATABASE, params.statusId, domain)
+	const statusText = await statusResponse.text()
+	if (!statusText) {
 		throw redirect(303, '/not-found')
 	}
-	return { status: JSON.parse(results), context: { ancestors: [], descendants: [] } }
+	const contextResponse = await contextAPI.handleRequest(domain, platform.DATABASE, params.statusId)
+	const contextText = await contextResponse.text()
+	const context = JSON.parse(contextText) as StatusContext
```
Comment (dario-piotrowicz) [inline frontend/src/routes/(frontend)/[accountId]/[statusId]/index.tsx:24]: I wonder what we should do if the context is not fetched successfully, any idea?

frontend/src/routes/(frontend)/[accountId]/[statusId]/index.tsx
```diff
@@ -4,20 +4,25 @@ import Status from '~/components/Status'
...
+>(async ({ request, redirect, platform, params }) => {
+	const domain = new URL(request.url).hostname
+	const statusResponse = await statusAPI.handleRequest(platform.DATABASE, params.statusId, domain)
+	const statusText = await statusResponse.text()
+	if (!statusText) {
 		throw redirect(303, '/not-found')
 	}
-	return { status: JSON.parse(results), context: { ancestors: [], descendants: [] } }
+	const contextResponse = await contextAPI.handleRequest(domain, platform.DATABASE, params.statusId)
+	const contextText = await contextResponse.text()
```
Comment (xtuc) [inline frontend/src/routes/(frontend)/[accountId]/[statusId]/index.tsx:23]:
```suggestion
	const contextText = await contextResponse.json()
```
can't you parse a JSON directly?

frontend/src/routes/(frontend)/[accountId]/[statusId]/index.tsx
Comment (xtuc) [reply frontend/src/routes/(frontend)/[accountId]/[statusId]/index.tsx:24]: I would throw an error because that shouldn't happen

Comment (petebacondarwin) [reply frontend/src/routes/(frontend)/[accountId]/[statusId]/index.tsx:24]: I think we can safely write an error to the logs and respond with `{ancestors: [], descendants: []}`

Comment (petebacondarwin) [reply frontend/src/routes/(frontend)/[accountId]/[statusId]/index.tsx:24]: LOL - or what @xtuc said.

petebacondarwin:
We should be aware that this could result in a large cascade of D1 queries + external fetches. Since each descendant then hits D1 again for the actor (which hopefully might be cached).

frontend/src/routes/(frontend)/[accountId]/[statusId]/index.tsx
Comment (petebacondarwin) [reply frontend/src/routes/(frontend)/[accountId]/[statusId]/index.tsx:23]: Sadly not. The `resp.json()` function returns a weird object that Qwik is not able to serialize.

Comment (xtuc) [reply frontend/src/routes/(frontend)/[accountId]/[statusId]/index.tsx:23]: I don't like that Qwik has those implicit things that don't work (same with state at the root or something like that) but ok.

xtuc (approved):

frontend/src/routes/(frontend)/[accountId]/[statusId]/index.tsx
Comment (dario-piotrowicz) [reply frontend/src/routes/(frontend)/[accountId]/[statusId]/index.tsx:24]: I guess I can just throw an error, so at least in the browser the user would get a failed network request and it will be slightly more evident that something went wrong

functions/api/v1/statuses.ts
```diff
@@ -128,5 +128,5 @@ export async function handleRequest(
 		...cors(),
 		'content-type': 'application/json; charset=utf-8',
 	}
-	return new Response(JSON.stringify(res), { headers })
+	return new Response(JSON.stringify({ ...res, mastodonId: note.mastodonId }), { headers })
```
xtuc, line 131: mastodonId is id already

xtuc (changes requested):

functions/api/v1/statuses.ts
dario-piotrowicz: I don't think so, I saw that it is different from the id we provide in the dummy data statuses and at each db setup it changes 😕

dario-piotrowicz: for example if I print the map in which I map the ids we provide in dummyData to mastodonIds I get this sort of result:
Screenshot 2023-01-18 at 14 22 11

Screenshot 2023-01-18 at 14 23 06

Screenshot 2023-01-18 at 14 23 31

as you can see our id is always the same but the mastodonId seems to change every time

dario-piotrowicz: anyways it doesn't matter as I am going to refactor the ui tests to use the lower level api there

frontend/src/routes/(frontend)/[accountId]/[statusId]/index.tsx
Comment (dario-piotrowicz) [reply frontend/src/routes/(frontend)/[accountId]/[statusId]/index.tsx:23]: yeah, unfortunately Qwik does create these sort of constraints 😓 

(although I think it is often understandable given the core idea of Qwik of splitting things up as much as possible)

frontend/src/routes/(frontend)/[accountId]/[statusId]/index.tsx
Comment (dario-piotrowicz) [reply frontend/src/routes/(frontend)/[accountId]/[statusId]/index.tsx]: I've put there a check for the content and a simple error throw, I thought of try-catching the other calls but it started to look overkill so I went with this very simple solution, let me know if that's ok 🙂

xtuc (approved):

frontend/mock-db/init.ts
```diff
@@ -1,57 +1,71 @@
...
+	return note
+}
+
+/**
+ * Reblogs a note (representing a status)
+ */
+async function reblogNote(db: D1Database, reblogger: Person, noteToReblog: Note) {
+	await addObjectInOutbox(db, reblogger, noteToReblog)
+	await insertReblog(db, reblogger, noteToReblog)
+}
```
petebacondarwin, line 43: I like the idea of avoiding the full external REST API here, especially given that we don't care about triggering notifications.

Could we move this to somewhere in the `backend/src/` project and reuse it here and in `functions/api/v1/statuses/[id]/reblog.ts`?

And similarly with other functions that are making assumptions about how the internals of the request handlers work?

petebacondarwin (approved):
Perhaps land this and the refactor to move these helpers to be shared with the production code?

frontend/mock-db/init.ts
dario-piotrowicz: ah yeah that's a good point 😃👍 

sure, I'll merge this if ok and then do some refactoring as a followup 🙂

dario-piotrowicz: PR opened with the refactoring 🙂  https://github.com/cloudflare/wildebeest/pull/151

Issue #147, opened 2023-01-21
CORS error: Method DELETE is not allowed by Access-Control-Allow-Methods in preflight response.


codehz:
So I cannot delete post from web client like https://pinafore.social/

PR #148, opened 2023-01-21
add DELETE method to cors allow-methods

codehz:
fix: #147

### DIFF
--- a/backend/src/utils/cors.ts
+++ b/backend/src/utils/cors.ts
@@ -2,6 +2,6 @@ export function cors(): object {
 	return {
 		'Access-Control-Allow-Origin': '*',
 		'Access-Control-Allow-Headers': 'content-type, authorization, idempotency-key',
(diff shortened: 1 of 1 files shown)


backend/src/utils/cors.ts
```diff
@@ -2,6 +2,6 @@ export function cors(): object {
 	return {
 		'Access-Control-Allow-Origin': '*',
 		'Access-Control-Allow-Headers': 'content-type, authorization, idempotency-key',
-		'Access-Control-Allow-Methods': 'GET, PUT, POST',
+		'Access-Control-Allow-Methods': 'GET, PUT, POST, DELETE',
```
petebacondarwin, line 5: Although this will enable the DELETE request to get past CORS, I don't believe that deletes are yet implemented in Wildebeest.

codehz: Ah, yes, it just failed silently...

PR #150, opened 2023-01-23
Copy edits

OutpostRyan:
Minor copy edits, including putting in a missing verb


petebacondarwin (approved):
Thanks for catching these!

PR #151, opened 2023-01-23
Reduce be fe code duplication

dario-piotrowicz:
Sharing some backend code with the frontend, so that we can use it during the db creation and remove code duplication, as suggested in https://github.com/cloudflare/wildebeest/pull/132#discussion_r1081117848


dario-piotrowicz:
I might have gone a bit overboard with the tsdoc comments 🤔

backend/src/activitypub/activities/handle.ts
```diff
@@ -287,16 +286,12 @@ export async function handle(
...
 
-			await Promise.all([
-				// Add the object in the originating actor's outbox, allowing other
-				// actors on this instance to see the note in their timelines.
-				addObjectInOutbox(db, fromActor, obj, activity.published),
-
-				// Store the reblog for counting
-				insertReblog(db, fromActor, obj),
-
-				sendReblogNotification(db, fromActor, targetActor, notifId, adminEmail, vapidKeys),
```
petebacondarwin, line 298: I think I would keep this function call separate. E.g.

```ts
await Promise.all([createReblog(...), sendReblogNotification(...)])
```

Sending a notification is not necessarily part of creating a reblog, hence the fact you had to make it optional.

backend/src/mastodon/reply.ts
```diff
@@ -1,4 +1,4 @@
-import type { Actor } from 'wildebeest/backend/src/activitypub/actors'
+import { Actor } from 'wildebeest/backend/src/activitypub/actors'
```
petebacondarwin, line 1: ??

backend/src/mastodon/status.ts
```diff
@@ -2,13 +2,16 @@ import type { Handle } from '../utils/parse'
...
+import { createPublicNote, type Note } from 'wildebeest/backend/src/activitypub/objects/note'
 import { loadExternalMastodonAccount } from 'wildebeest/backend/src/mastodon/account'
 import * as actors from 'wildebeest/backend/src/activitypub/actors'
 import * as media from 'wildebeest/backend/src/media/'
 import type { MastodonStatus } from 'wildebeest/backend/src/types'
 import { parseHandle } from '../utils/parse'
 import { urlToHandle } from '../utils/handle'
+import type { Person } from 'wildebeest/backend/src/activitypub/actors'
+import { addObjectInOutbox } from '../activitypub/actors/outbox'
+import type { Object as ActivityPubObject } from 'wildebeest/backend/src/activitypub/objects'
```
petebacondarwin, line 14: In a future PR we should just rename `Object` to `APObject`.

petebacondarwin (approved):
One suggestion...

backend/src/mastodon/reply.ts
dario-piotrowicz: sorry I missed this one 😓

backend/src/mastodon/status.ts
dario-piotrowicz: yeah 100% agree, `Object` is extremely ambiguous and also the vscode fails to auto-import it (since it doesn't see a missing type)

I would likely go with `ActivityPubObject` I think, but I tend to be rather verbose with my naming 😁 😬

backend/src/activitypub/activities/handle.ts
dario-piotrowicz: yeah good point 🙂👍

PR #152, opened 2023-01-24
Images grid

dario-piotrowicz:
Adding a media grid to display the media attachments (instead of showing them one after the other)

Result with 1,2,3 and 4 images:

Screenshot 2023-01-24 at 11 25 10

Screenshot 2023-01-24 at 11 24 27

Screenshot 2023-01-24 at 11 23 46

Screenshot 2023-01-24 at 11 09 56

___

As part of this change I improved the video element, before it wasn't showing the controls, I've also added a black background

before:
Screenshot 2023-01-24 at 11 56 11

after:
Screenshot 2023-01-24 at 11 21 55

___

The video are also included in the grid, for example this is a grid with 3 images and one video:

Screenshot 2023-01-24 at 11 22 07

___

> Notes:
> - I think that the image modal should be improved I can do that in a followup PR
> - Audios are not yet implemented so that are not included here
> - The video needs a black background and can't stretch like the images because otherwise the controls would get cropped out

### DIFF
--- a/backend/src/activitypub/objects/note.ts
+++ b/backend/src/activitypub/objects/note.ts
@@ -25,7 +25,7 @@ export async function createPublicNote(
 	db: D1Database,
 	content: string,
 	actor: Actor,
-	attachment: Array<objects.APObject> = [],
+	attachments: Array<objects.APObject> = [],
 	extraProperties: any = {}
 ): Promise<Note> {
 	const actorId = new URL(actor.id)
@@ -41,8 +41,8 @@ export async function createPublicNote(
 		sensitive: false,
 		summary: null,
 		tag: [],
-		attachment,
 
+		attachment: attachments,
 		inReplyTo: null,
 		...extraProperties,
 	}
--- a/backend/src/media/index.ts
+++ b/backend/src/media/index.ts
@@ -6,6 +6,8 @@ import type { APObject } from 'wildebeest/backend/src/activitypub/objects'
 export function fromObject(obj: APObject): MediaAttachment {
 	if (obj.type === IMAGE) {
 		return fromObjectImage(obj)
+	} else if (obj.type === 'Video') {
+		return fromObjectVideo(obj)
 	} else if (obj.type === 'Document') {
 		return fromObjectDocument(obj as Document)
 	} else {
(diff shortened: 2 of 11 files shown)


backend/src/activitypub/objects/note.ts
```diff
@@ -25,7 +25,7 @@ export async function createPublicNote(
 	db: D1Database,
 	content: string,
 	actor: Actor,
-	attachment: Array<objects.Object> = [],
+	attachments: Array<objects.Object> = [],
```
dario-piotrowicz, line 28: not an important change but I found `attachment` to be a misleading

backend/src/media/index.ts
```diff
@@ -4,8 +4,10 @@ import { IMAGE } from 'wildebeest/backend/src/activitypub/objects/image'
 import type { Object } from 'wildebeest/backend/src/activitypub/objects'
 
 export function fromObject(obj: Object): MediaAttachment {
-	if (obj.type === IMAGE) {
+	if (IMAGE.toLowerCase() === obj.type.toLowerCase()) {
```
dario-piotrowicz, line 7: In our dummy data and from mastodon the type is `image` but here we were only handling `Image`... not sure what's up with this 😕 🤷

frontend/src/components/MediaGallery.tsx/index.scss
```diff
@@ -0,0 +1,34 @@
+.media-gallery:has(:nth-child(1)) {
```
dario-piotrowicz, line 1: here I thought that using `:has` to check the number of children and change the grid accordingly was a cool solution, but this becomes the only code (except from the dynamic statuses contents) that is not using tailwind, so I am a bit torn on whether I should just change this to use tailwind.

The issue is that tailwind doesn't provide functional pseudo-classes (see this [tweet](https://twitter.com/adactio/status/1404467229857665027?lang=en)) so if I were to use tailwind I would have to have code that checks the number of medias and apply different classes accordingly.
I would also have to take into consideration the children and so something analogous for them...

I feel like that code would end up being a bit messy... but I can give it a shot if we think that would be the right direction?

any thoughts @jculvey @petebacondarwin ?

petebacondarwin:
Love the screenshots!!

backend/src/media/index.ts
xtuc: It should be capitalized Image here because it's an ActivityOub object. Mastodon uses lower case image as media type in their API but should use that code path. We need a convert first

xtuc (changes requested):

backend/src/activitypub/objects/note.ts
```diff
@@ -41,8 +41,8 @@ export async function createPublicNote(
 		sensitive: false,
 		summary: null,
 		tag: [],
-		attachment,
 
+		attachment: attachments,
```
dario-piotrowicz, line 45: @xtuc in your comment: https://github.com/cloudflare/wildebeest/pull/152/files#r1085391616
are you saying that here we should take the attachments and replace all types `image` and `video` with `Image` and `Video`? (`audio` and `Audio` too I guess?)

frontend/src/components/MediaGallery.tsx/Image.tsx
```diff
@@ -5,26 +5,11 @@ type Props = {
 	mediaAttachment: MediaAttachment
 }
 
-export const focusToObjectFit = (focus: { x: number; y: number }) => {
```
jculvey, line 8: This bit actually does something important -- given a user provided focal point, it will zoom the preview image to that location. 

Visually, with this focus logic:
image

without:
image

frontend/src/components/MediaGallery.tsx/index.scss
jculvey: I would stick with this class you've crafted if it works.

jculvey (approved):

backend/src/activitypub/objects/note.ts
dario-piotrowicz: Not applied as it is not the right solution

backend/src/media/index.ts
dario-piotrowicz: reverted, I'm converting our dummyData statuses instead

frontend/src/dummyData.tsx
```diff
@@ -1776,6 +1777,24 @@ export const statuses: MastodonStatus[] = [
 	},
 ]
 
+export const statuses: MastodonStatus[] = mastodonRawStatuses.map((rawStatus) => ({
```
dario-piotrowicz, line 1780: @jculvey are you ok with this? 🙂

PR #154, opened 2023-01-24
rename ambiguous Object to APObject

dario-piotrowicz:
As suggested here: https://github.com/cloudflare/wildebeest/pull/151#discussion_r1084453384


backend/src/mastodon/reblog.ts
```diff
@@ -12,11 +12,11 @@ import { addObjectInOutbox } from '../activitypub/actors/outbox'
  * @param actor Reblogger
  * @param obj Object to reblog
```
petebacondarwin, line 13: Perhaps: `@param obj ActivityPub object to reblog`?

PR #163, opened 2023-01-31
Add oauth state param

xmflsct:
Help preventing potential CSRF https://auth0.com/docs/secure/attack-protection/state-parameters. Also make it work for the clients that do check the `state` param.

### DIFF
--- a/backend/src/mastodon/timeline.ts
+++ b/backend/src/mastodon/timeline.ts
@@ -1,5 +1,4 @@
 import type { MastodonStatus } from 'wildebeest/backend/src/types/status'
-import { getFollowingId } from 'wildebeest/backend/src/mastodon/follow'
 import type { Actor } from 'wildebeest/backend/src/activitypub/actors/'
 import { toMastodonStatusFromRow } from './status'
 import { PUBLIC_GROUP } from 'wildebeest/backend/src/activitypub/activities'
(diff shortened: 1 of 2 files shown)


backend/src/errors/index.ts
```diff
@@ -34,6 +34,10 @@ export function clientUnknown(): Response {
 	return generateErrorResponse(`The client is unknown or invalid`, 403)
 }
 
+export function notFound(): Response {
```
xtuc, line 37: Could you please remove the 404 change from this PR? I think it needs a change

functions/oauth/authorize.ts
```diff
@@ -73,11 +76,11 @@ export async function handleRequestPost(
 	if (person === null) {
 		url.pathname = '/first-login'
 		url.searchParams.set('email', identity.email)
-		url.searchParams.set('redirect_uri', encodeURIComponent(redirect_uri + '?code=' + code))
+		url.searchParams.set('redirect_uri', encodeURIComponent(redirect_uri + '?code=' + code + state))
 		return URLsafeRedirect(url.toString())
 	}
 
-	return URLsafeRedirect(redirect_uri + '?code=' + code)
+	return URLsafeRedirect(redirect_uri + '?code=' + code + state)
```
dario-piotrowicz, line 83: we've got two `redirect_uri + '?code=' + code + state`, wouldn't it be better to define a variable on line 74 for it and reuse it in both places? 

like 
```
const redirectUriParam = `redirect_uri?code=${code}${state}`
```
?

backend/src/errors/index.ts
xmflsct: Yep sorry that was not intentional.

functions/oauth/authorize.ts
xmflsct: Checked

PR #170, opened 2023-02-01
Add `tag` on note

xtuc:


backend/src/mastodon/status.ts
```diff
@@ -25,7 +26,15 @@ export function getMentions(input: string): Array<Handle> {
 				i++
 			}
 
-			mentions.push(parseHandle(buffer))
+			const handle = parseHandle(buffer)
+			const domain = handle.domain ? handle.domain : instanceDomain
+			const acct = `${handle.localPart}@${domain}`
+			const targetActor = await queryAcct(domain!, acct)
```
dario-piotrowicz, line 32: is the non-null assertion needed? (isn't domain always going to be defined?)
```suggestion
			const targetActor = await queryAcct(domain, acct)
```

PR #172, opened 2023-02-01
simplify enrichStatus method by using regexes

dario-piotrowicz:
as I mentioned in https://github.com/cloudflare/wildebeest/pull/169 I felt like regexes could help making the solution simpler and more concise so I gave it a shot 🙂

### DIFF
--- a/backend/src/mastodon/microformats.ts
+++ b/backend/src/mastodon/microformats.ts
@@ -9,66 +9,38 @@ function tag(name: string, content: string, attrs: Record<string, string> = {}):
 	return `<${name}${htmlAttrs}>${content}</${name}>`
 }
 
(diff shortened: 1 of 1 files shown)


xtuc:
Curious what @petebacondarwin thinks. I'm usually not a fan of clever regexes because they aren't flexible / maintainable, we might in the future have more complex logic.

petebacondarwin:
In this case I agree that the regex is simpler here than the hand coded algorithm.
Either way, the real key is to have comprehensive tests. I would be worried to update either approach without them.
FWIW Mastodon uses regexes too: https://github.com/mastodon/mastodon/blob/725f21662f7ba287b36adc1d973c2aa57296c781/app/lib/extractor.rb#L25

xtuc (approved):

PR #182, opened 2023-02-03
implement account page

dario-piotrowicz:
Result:
Screenshot 2023-02-03 at 12 22 04

(Note: this PR is for the basic info only, I'll look into the other info in a followup PR)

<details>
<summary>Mastodon</summary>

Screenshot 2023-02-03 at 12 09 49
</details>

### DIFF
--- a/backend/src/activitypub/objects/index.ts
+++ b/backend/src/activitypub/objects/index.ts
@@ -231,7 +231,7 @@ export async function sanitizeObjectProperties(properties: unknown): Promise<APO
  * See https://docs.joinmastodon.org/spec/activitypub/#sanitization
  */
 export async function sanitizeContent(unsafeContent: string): Promise<string> {
-	return await contentRewriter.transform(new Response(unsafeContent)).text()
+	return await getContentRewriter().transform(new Response(unsafeContent)).text()
 }
 
 /**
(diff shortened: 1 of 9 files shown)


functions/api/v1/accounts/[id].ts
```diff
@@ -21,41 +16,11 @@ export const onRequest: PagesFunction<Env, any, ContextData> = async ({ request,
...
-		// Retrieve the statuses from a local user
-		return getLocalAccount(domain, db, handle)
-	} else if (handle.domain !== null) {
-		// Retrieve the statuses of a remote actor
-		const acct = `${handle.localPart}@${handle.domain}`
-		return getRemoteAccount(handle, acct)
+	if (account) {
+		return new Response(JSON.stringify(account), { headers })
 	} else {
-		return new Response('', { status: 403 })
```
Comment (dario-piotrowicz) [inline functions/api/v1/accounts/[id].ts:34]: @xtuc you can notice that the 403 is gone in my code, I think that it was never anyway returned, because if `handle.domain` was `null` then you'd go in the first branch, if it was different from `null` you'd either go in the first branch or the second, so I don't think the last else was actually doing anything

please let me know if I'm wrong 🙂

backend/src/activitypub/objects/index.ts
```diff
@@ -231,7 +231,7 @@ export async function sanitizeObjectProperties(properties: unknown): Promise<APO
  * See https://docs.joinmastodon.org/spec/activitypub/#sanitization
  */
 export async function sanitizeContent(unsafeContent: string): Promise<string> {
-	return await contentRewriter.transform(new Response(unsafeContent)).text()
+	return await getContentRewriter().transform(new Response(unsafeContent)).text()
```
petebacondarwin, line 234: Out of curiosity, why now create a new rewriter every time?

frontend/src/routes/(frontend)/[accountId]/index.tsx
```diff
@@ -1,23 +1,131 @@
...
+		},
+	]
+
+	const accountDomain = getAccountDomain(accountDetails)
+
+	return (
+		<div>
+			<StickyHeader>
+				<div class="flex justify-between items-center xl:rounded-t header bg-wildebeest-700">
+					<button class="text-semi no-underline text-wildebeest-vibrant-400 bg-transparent p-4" onClick$={goBack}>
```
Comment (petebacondarwin) [inline frontend/src/routes/(frontend)/[accountId]/index.tsx:78]: Why is this a button and not an anchor?

Comment (petebacondarwin) [reply frontend/src/routes/(frontend)/[accountId]/index.tsx:78]: Oh it be because you want to literally do "back"...

petebacondarwin (approved):
Awesome. Love it!

backend/src/activitypub/objects/index.ts
xtuc: yeah we shouldn't need.

functions/api/v1/accounts/[id].ts
Comment (xtuc) [reply functions/api/v1/accounts/[id].ts:34]: Yeah that's correct. The response was mostly for typing here. Thanks for checking

xtuc (approved):

backend/src/activitypub/objects/index.ts
dario-piotrowicz: ah yeah, good spot! 😄 .... I was thinking that this might need a comment 😅 (but I didn't just to keep the code cleaner 😓)

The problem is that we import backend stuff from the frontend and this file is getting imported as well (for the account page). I was getting the error that HTMLRewriter is not defined, since it is not in the browser, so I wrapped the creation in these functions so that they HTMLRewriter gets created only when needed (so only in the backend) solving the issue

___

Actually, we shouldn't import stuff from the backend right?

We do it in a handful of places:
Screenshot 2023-02-03 at 13 44 49

I think it would be great if we could avoid that, maybe we can create a shared library which backend and frontend can import both from? 🤔

frontend/src/routes/(frontend)/[accountId]/index.tsx
Comment (dario-piotrowicz) [reply frontend/src/routes/(frontend)/[accountId]/index.tsx:78]: yeah exactly 🙂 

actually I am tempted to apply this change to the other back anchors as well, since I notice that those point to `/explore` which I feel like may not be the right location in some situations

PR #190, opened 2023-02-03
add error page and redirect to such page in the various loader$s

dario-piotrowicz:
Result (in case there in an error, in `/explore` for example):
Screenshot 2023-02-03 at 23 23 01


xtuc (approved):

frontend/src/utils/getNotFoundHtml/not-found-raw.html
```diff
@@ -30,7 +30,7 @@
     </style>
 </head>
 <body>
-		<div>
+		<main>
 			<h1>The page you are looking for isn't here.</h1>
 		</div>
```
xtuc, line 35:
```suggestion
		</main>
```

PR #199, opened 2023-02-06
try to fix playwright actions

dario-piotrowicz:


dario-piotrowicz:
No longer needed (I think)

PR #212, opened 2023-02-07
MOW-133: add idempotency for posting

xtuc:


backend/src/mastodon/idempotency.ts
```diff
@@ -0,0 +1,51 @@
...
+        INNER JOIN objects ON objects.id = idempotency_keys.object_id
+        WHERE idempotency_keys.key = ?1 AND expires_at >= datetime() 
+    `
+
+	const { results, success, error } = await db.prepare(query).bind(key).all<any>()
+	if (!success) {
+		throw new Error('SQL error: ' + error)
+	}
+
+	if (!results || results.length === 0) {
```
dario-piotrowicz, line 33: `!result` here is superfluous since you're already checking that and throwing above

backend/src/mastodon/idempotency.ts
```diff
@@ -0,0 +1,51 @@
...
+		return null
+	}
+
+	const result: any = results[0]
```
dario-piotrowicz, line 37: I don't think specifying the type `any` is needed?

backend/src/mastodon/idempotency.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export async function hasKey(db: D1Database, key: string): Promise<APObject | null> {
+	const query = `
+        SELECT objects.*
+        FROM idempotency_keys
+        INNER JOIN objects ON objects.id = idempotency_keys.object_id
+        WHERE idempotency_keys.key = ?1 AND expires_at >= datetime() 
+    `
+
+	const { results, success, error } = await db.prepare(query).bind(key).all<any>()
```
dario-piotrowicz, line 28: could we use anything better instead of `any`?

backend/test/mastodon/statuses.spec.ts
```diff
@@ -875,5 +874,44 @@ describe('Mastodon APIs', () => {
...
+					headers: {
+						'content-type': 'application/json',
+						'idempotency-key': idempotencyKey,
+					},
+					body: JSON.stringify(body),
+				})
+
+			const res1 = await statuses.handleRequest(req(), db, actor, userKEK, queue, cache)
+			assert.equal(res1.status, 200)
+			const data1 = await res1.json<any>()
```
dario-piotrowicz, line 898: is the `<any>` needed?

backend/src/mastodon/idempotency.ts
xtuc: I wish we could generate table typings as TS from D1, would be easier than manually typing the object table.

dario-piotrowicz: yeah 😕

dario-piotrowicz (approved):

backend/src/mastodon/idempotency.ts
xtuc: Well let's do it. It's useful for D1

PR #221, opened 2023-02-08
adjust account urls

### LINKED ISSUE #216: Bug: wrong URLs used in the account name link
For an account we have two different urls one used on the avatar and one on the name link, they are different and only one currently works (locally only the name link works and in prod only the avatar one does) 
Screenshot 2023-02-07 at 17 28 11

The two urls need to be the same and it should be:
 - a full remote url for users registered on other servers
 - a relative url (allowing spa navigation) for users registered on the instance

dario-piotrowicz:
resolves #216


backend/src/accounts/adjustLocalHostDomain.ts
```diff
@@ -0,0 +1,12 @@
...
+ * in that case replaces it with '0.0.0.0' (which is what we use for our local data)
+ *
+ * Note: only needed for local development
+ *
+ * @param domain the potentially localhost domain
+ * @returns the adjusted domain if it was a localhost one, the original domain otherwise
+ */
+export function adjustLocalHostDomain(domain: string) {
+	return domain.replace(/^localhost$|^127(\.(?:\d){1,3}){3}$/, '0.0.0.0')
+}
```
xtuc, line 12: I guess a better place would be in utils

dario-piotrowicz: good point 🙂👍

xtuc (approved):

PR #226, opened 2023-02-08
update the logos use to cloudflare.social

dario-piotrowicz:


dario-piotrowicz:
I misunderstood the requirement here, sorry about that 🙂

Issue #240, opened 2023-02-09
Unable to log in


koehn:
After much mucking about, I was able to install the application, created a login policy, and I can see the main page. 

When I try to log in using an app (I tried pinafore, the Mastodon iOS app, and Ivory):
1. I enter my email address, corresponding to the policy I created
2. I enter the validation code that gets sent to my email

At this point, I get a blank page with a URL like `https://dogfood.social/oauth/authorize?client_id=[redacted]&redirect_uri=https%3A%2F%2Fpinafore.social%2Fsettings%2Finstances%2Fadd&response_type=code&scope=read%20write%20follow%20push`. 

If I View Source from a browser, the source of the login page is empty (no HTML at all). 

In the logs I see the following:

```
{
  "outcome": "ok",
  "scriptName": "pages-worker--844423-production",
  "exceptions": [],
  "logs": [
    {
      "message": [
        "-> GET https://dogfood.social/oauth/authorize?client_id=[redacted]&redirect_uri=https%3A%2F%2Fpinafore.social%2Fsettings%2Finstances%2Fadd&response_type=code&scope=read%20write%20follow%20push "
      ],
      "level": "log",
      "timestamp": 1675971779529
    },
    {
      "message": [
        "<- 200"
      ],
      "level": "log",
      "timestamp": 1675971779652
    },
    {
      "message": [
        "Error: invalid handle: localPart: https://dogfood.social/ap/users/brad"
      ],
      "level": "error",
      "timestamp": 1675971779652
    }
  ],
  "eventTimestamp": 1675971779529,
  "event": {
    "request": {
      "url": "https://dogfood.social/oauth/authorize?client_id=REDACTED&redirect_uri=https%3A%2F%2Fpinafore.social%2Fsettings%2Finstances%2Fadd&response_type=code&scope=read%20write%20follow%20push",
      "method": "GET",
      "headers": {
        "accept": "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8",
        "accept-encoding": "gzip",
        "accept-language": "en-US,en;q=0.9",
        "cf-access-jwt-assertion": "REDACTED",
        "cf-connecting-ip": "2606:54c0:7681:27f0::e:100",
        "cf-connecting-o2o": "1",
        "cf-ipcountry": "US",
        "cf-ray": "796f1765fb67e20e",
        "cf-visitor": "{\"scheme\":\"https\"}",
        "connection": "Keep-Alive",
[...]

PR #241, opened 2023-02-09
Parse handle

koehn:
Attempts to fix a bug #240 where parsing URI handles failed.


koehn:
Closing because reverting the change *also* fixes the problem.

PR #250, opened 2023-02-10
fix wrong account url being used

dario-piotrowicz:


frontend/src/components/Status/index.tsx
```diff
@@ -17,7 +18,7 @@ export default component$((props: Props) => {
 	const status = props.status.reblog ?? props.status
 	const reblogger = props.status.reblog && props.status.account
 
-	const accountUrl = `/@${status.account.username}`
+	const accountUrl = useAccountUrl(status.account)
```
dario-piotrowicz, line 21: in my original change I forgot to update this one
(https://github.com/cloudflare/wildebeest/blob/c8e9f04b9b77e64f562dea52bf991a149e36bf89/frontend/src/components/Status/index.tsx#L21)

I wonder if that was the only issue

53jk1 (approved):

PR #251, opened 2023-02-10
Fix for cloudflare#248

koehn:
Enables GIF and webp images in addition to jpeg and png.

### DIFF
--- a/backend/src/media/index.ts
+++ b/backend/src/media/index.ts
@@ -16,8 +16,9 @@ export function fromObject(obj: APObject): MediaAttachment {
 	}
 }
 
+const imageTypes = ['image/jpeg', 'image/png', 'image/gif', 'image/webp']
 export function fromObjectDocument(obj: Document): MediaAttachment {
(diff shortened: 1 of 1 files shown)


dario-piotrowicz (approved):

backend/src/media/index.ts
```diff
@@ -16,8 +16,9 @@ export function fromObject(obj: APObject): MediaAttachment {
 	}
 }
 
+const imageTypes = ['image/jpeg', 'image/png', 'image/gif', 'image/webp']
 export function fromObjectDocument(obj: Document): MediaAttachment {
-	if (obj.mediaType === 'image/jpeg' || obj.mediaType === 'image/png') {
+	if (imageTypes.includes(obj.mediaType as string)) {
 		return fromObjectImage(obj)
 	} else if (obj.mediaType === 'video/mp4') {
```
dario-piotrowicz, line 23: shall we do something like this for videos as well?

PR #272, opened 2023-02-13
introduce basic for hashtag

xtuc:
This change detects hashtags in statuses and insert them into a new table `note_hashtags`.


dario-piotrowicz (approved):
nice 😃

PR #283, opened 2023-02-13
Improve infinite scrolling

dario-piotrowicz:
Add infinite scrolling to local and federated timeline pages

Add infinite scrolling to an account page as well

Add e2e tests for the infinite scrolling of all the above + the existing explore page


CyberFlameGO (approved):
Looks like it works!

ui-e2e-tests/infinite-scrolling.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+	)
+})
+
+/**
+ * generates a function that creates mock statuses when called,
+ * it uses a closure to keep track of the number of generated
+ * statuses to that they are consistently enumerated
+ * ('Mock Fetched Status #000', 'Mock Fetched Status #001', ...)
+ */
+export function getMockStatusFn(): () => MastodonStatus {
```
dario-piotrowicz, line 84: This should instead use the utilities defined in https://github.com/cloudflare/wildebeest/pull/285

We should merge that PR first and update the code here afterwards before merging

dario-piotrowicz:
I've converted the PR to draft just to make sure we don't merge it before https://github.com/cloudflare/wildebeest/pull/285

ui-e2e-tests/infinite-scrolling.spec.ts
dario-piotrowicz: For some reason playwright breaks when I try to import those utilities... :disappointed: 

I've at least updated this code to also not include real data, hopefully it should be good enough

xtuc (approved):

PR #309, opened 2023-02-17
/api/v1/statuses should be public for public statuses

celso:


backend/src/middleware/main.ts
```diff
@@ -49,6 +49,7 @@ export async function main(context: EventContext<Env, any, any>) {
 	if (
 		url.pathname === '/oauth/token' ||
 		url.pathname === '/oauth/authorize' || // Cloudflare Access runs on /oauth/authorize
+		url.pathname === '/api/v1/statuses' || // Unless private https://docs.joinmastodon.org/methods/statuses/#get
```
xtuc, line 52: Should be `url.pathname.startsWith('/api/v1/statuses/')` no?

xtuc: Ideally a regex to keep `/api/v1/statuses/[id]/reblog` or  `/api/v1/statuses/[id]/favourite` authentificated

celso: Makes sense, added the regex

xtuc (approved):

PR #310, opened 2023-02-17
add CI=true to database:create-mock

dario-piotrowicz:
add CI=true to database:create-mock so that wrangler doesn't show a confirmation prompt asking the user if they want to apply the migrations (since at that point it is obvious they do)


dario-piotrowicz:
just fixing a very very small thing that's been annoying me 😛

PR #312, opened 2023-02-18
fix: Changed line breaks so that text does not overflow

yug1224:
Thanks for the great project!

Changed the display of line breaks and spaces.

|  | Before | After |
| :--: | :--: | :--: |
| xl | xl-before | xl-after |
| md | md-before | md-after |
| sm | sm-before | sm-after |

PR #318, opened 2023-02-20
remove tabs and newlines from dummy status

dario-piotrowicz:
As indicated in https://github.com/cloudflare/wildebeest/pull/312 statuses content can include spacing so this PR fixes one of our dummy statuses so that it doesn't contain extra spacing

before:
Screenshot 2023-02-20 at 10 39 01

after:
Screenshot 2023-02-20 at 10 37 54

### DIFF
--- a/frontend/src/dummyData/statuses.ts
+++ b/frontend/src/dummyData/statuses.ts
@@ -8,6 +8,7 @@ const mastodonRawStatuses: MastodonStatus[] = [
 		content: `
 		<p>Fine. I'll use Wildebeest!</p>
 		<p>It does look interesting:
+			&#32
 			<a href="https://blog.cloudflare.com/welcome-to-wildebeest-the-fediverse-on-cloudflare/"
 				target="_blank"
 				rel="nofollow noopener noreferrer">
@@ -15,7 +16,7 @@ const mastodonRawStatuses: MastodonStatus[] = [
 					<span class="ellipsis">blog.cloudflare.com/welcome-to</span>
 					<span class="invisible">-wildebeest-the-fediverse-on-cloudflare/</span>
 			</a>
-		</p>`,
+		</p>`.replace(/[\t\n]/g, ''),
 		account: george,
 	}),
 	generateDummyStatus({


frontend/src/dummyData/statuses.ts
```diff
@@ -8,14 +8,14 @@ const mastodonRawStatuses: MastodonStatus[] = [
 		`
 		<p>Fine. I'll use Wildebeest!</p>
 		<p>It does look interesting:
-			<a href="https://blog.cloudflare.com/welcome-to-wildebeest-the-fediverse-on-cloudflare/"
+			 <a href="https://blog.cloudflare.com/welcome-to-wildebeest-the-fediverse-on-cloudflare/"
```
petebacondarwin, line 11: ??

petebacondarwin:
I feel a bit weird that we rendering these whitespace characters, this is not how HTML should work IMO.
Is this what other ActivityPub clients do?

frontend/src/dummyData/statuses.ts
dario-piotrowicz: 😅 

yeah I am removing all tabs and newlines, and I did want a space between `interesting:` and the link

would the space on line 10 be better? and/or also moving the `<a` there?

dario-piotrowicz:
> I feel a bit weird that we rendering these whitespace characters, this is not how HTML should work IMO. Is this what other ActivityPub clients do?

yeah I totally agree, but I did experiment with mastodon and they do keep the spacing 😕 (so I guess we should as well? 🤷 )
Screenshot at 2023-02-20 10-54-40

frontend/src/dummyData/statuses.ts
petebacondarwin: Hmm. So in that case I think you should be replacing the whitespace below with a space, not an empty string.

petebacondarwin: Or if you really want to do it here (and maintain the current replacement behaviour) then I think you should use an HTML encoding: `&#32`.

dario-piotrowicz: > Hmm. So in that case I think you should be replacing the whitespace below with a space, not an empty string.
the issue is that I am clearing up all the tabs and newlines, it's not like I am contracting them or the like, so I don't think I could change the replacement to keep a tab (unless I don't replace the specific tab found after `interesting:` but that doesn't really look clean 😕)

dario-piotrowicz: > Or if you really want to do it here (and maintain the current replacement behaviour) then I think you should use an HTML encoding: &#32.
cool thanks I'll use that 😃

petebacondarwin: > the issue is that I am clearing up all the tabs and newlines, it's not like I am contracting them or the like, so I don't think I could change the replacement to keep a tab (unless I don't replace the specific tab found after interesting: but that doesn't really look clean 😕)

Not sure if we are on the same page here. I was suggesting that you replace all the tabs and newlines with a space rather than an empty string.  E.g. `.replace(/[\t\n]/g, ' ')`. I think this should have the same effect but not cause text span to run into each other, right?

dario-piotrowicz: I see 🙂 

Anyways that wouldn't work as we have multiple tab spaces, so I tried `.replace(/[\t\n]+/g, ' ')` it still produces a bit more spaces than necessary, so I tried concatenating `.replace(/\s+/g, ' ')` to squash those but then I realized that we could have spaces before and after tags which aren't really consecutive but produce a result with too many spaces anyways, since this is only for a mock status I actually ended up just using `&#32` as you suggested

if you do prefer not to have it and use replacements alone I can also use the above ones and have a small number of extra spaces, either way this is just to show our mock content so I don't think it is too important, nor it needs to be precise

Let me know if `&#32` is ok or if you'd prefer a different solution 🙂

petebacondarwin (approved):
Let's go with this. As you say it is only test data.
But I do feel very weird about the `whitespace: pre-wrap` display of toots... It seems to me that clients should provide proper wysiwyg editing that generates HTML if that is what you want - or HTML should not be allowed at all.

dario-piotrowicz:
> Let's go with this. As you say it is only test data. But I do feel very weird about the `whitespace: pre-wrap` display of toots... It seems to me that clients should provide proper wysiwyg editing that generates HTML if that is what you want - or HTML should not be allowed at all.

Cool thanks 😃

Regarding the pre-wrap again I totally agree with you it does feel quite weird, but Mastodon seems to be applying it and we've been following Mastodon's implementation for all our other decisions ... 🤷 ?

Issue #282, opened 2023-02-13
Support for Instance About Page


bretcarmichael:
New Wildebeest instances will benefit from having About pages that can be managed by the instance admin. The feature should provide the following capabilities:

- Administered by
- - Who administers the server and their contact information. This is needed in case an instance user or federated user has a complaint about content on the server that might violate a moderation policy. Absent this, other instances will just start blacklisting  given instances that are not aware of the need for action. The information is also needed for DMCA takedown requests and to facilitate S230 protections.
- About
- - This is just a topical description of the instance.
- Server Rules
- - An list detailing content that is allowed and disallowed on the instance. Absent this, there are added moderated challenges and possible legal challenges.
- Moderated Servers
- - This is a list of blocked servers. Fediverse servers can have a multitude of content types, and admins need to protect their uses and reduce the risk of lawsuits. Admins need the capability to blacklist other instances that allow illegal and unwanted content , including but not limited to: harassment, bigotry, racism, sexism, sexual exploitation, antisemitism, etc.

nthomasgsy:
Boost: this is probably the most suitable way to achieve having somewhere to link to a Privacy Policy as I was requesting in closed issue https://github.com/cloudflare/wildebeest/issues/281.  This is the same point as @bretcarmichael second bullet above and should be treated as a serious issue.

nthomasgsy:
@celso thanks for your comment on the other issue. This proposition for /about would be a good, standard, way of getting at least part way there.

celso:
We're looking into this. More news soon.

bretcarmichael:
Awesome. Thank you! 😀

PR #325, opened 2023-02-20
implement UI for about page

dario-piotrowicz:
initial step for #282

___
__Note__: Currently it uses data I just copy pasted from Mastodon, but it is not accessible by the user since the page's loader redirects to `/`
___

Result:
Screenshot at 2023-02-20 18-04-28

Screenshot at 2023-02-20 18-05-02

### DIFF
--- a/frontend/src/routes/(frontend)/about/index.tsx
+++ b/frontend/src/routes/(frontend)/about/index.tsx
@@ -1,0 +1,142 @@
+import { component$ } from '@builder.io/qwik'
+import { DocumentHead, loader$ } from '@builder.io/qwik-city'
+import { getDomain } from 'wildebeest/backend/src/utils/getDomain'
+import { Accordion } from '~/components/Accordion/Accordion'
+import { AccountCard } from '~/components/AccountCard/AccountCard'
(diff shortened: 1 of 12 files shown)


xtuc (approved):
lgtm if the page is hidden / not accessible at the moment

frontend/src/routes/(frontend)/about/index.tsx
```diff
@@ -0,0 +1,142 @@
...
+	rules: { id: string; text: string }[]
+	extended_description: {
+		updated_at: string
+		content: string
+	}
+}
+
+export const aboutInfoLoader = loader$<Promise<AboutInfo>>(async ({ resolveValue, request, redirect }) => {
+	// TODO: properly implement loader and remove redirect
+	throw redirect(302, '/')
```
dario-piotrowicz, line 28: @xtuc 🙂

dario-piotrowicz:
> lgtm if the page is hidden / not accessible at the moment

yeah, as you can see here: https://github.com/cloudflare/wildebeest/pull/325/files#r1112297297 we just always unconditionally redirect to `/` when you try to access the page 🙂

PR #328, opened 2023-02-22
remove break-all

dario-piotrowicz:
This PR fixes text getting broken up too dynamically, resulting in weird views like these:
Screenshot 2023-02-22 at 12 42 29

Screenshot 2023-02-22 at 12 43 40

____

before:
before

after:
after

### DIFF
--- a/frontend/src/components/AccountCard/AccountCard.tsx
+++ b/frontend/src/components/AccountCard/AccountCard.tsx
@@ -13,18 +13,17 @@ export const AccountCard = component$<{
 	const accountUrl = useAccountUrl(account)
 
 	return (
-		<Link
-			href={accountUrl}
-			class="inline-grid grid-cols-[repeat(2,_max-content)] grid-rows-[1fr,1fr] items-center no-underline"
-		>
-			<div class="row-span-2">
+		<Link href={accountUrl} class="inline-flex items-center no-underline flex-wrap gap-2">
+			<div class="flex-grow flex-shrink-0 flex justify-center">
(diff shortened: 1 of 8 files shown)


dario-piotrowicz:
@yug1224 pinging you in case you want to have a look and see if the fix works for you (by the way I hope you don't mind I reused your test content)

frontend/src/components/AccountCard/AccountCard.tsx
```diff
@@ -13,18 +13,17 @@ export const AccountCard = component$<{
...
+			<div class="flex-grow flex-shrink-0 flex justify-center">
 				<Avatar primary={account} secondary={secondaryAvatar ?? null} />
 			</div>
-			<div data-testid="account-display-name" class="ml-2 col-start-2 row-start-1">
-				{getDisplayNameElement(account)}
-			</div>
-			<div class="ml-2 text-wildebeest-400 col-start-2 row-start-2">
-				@{subText === 'username' ? account.username : account.acct}
+			<div>
+				<div data-testid="account-display-name" class="col-start-2 row-start-1">
```
petebacondarwin, line 21: Do the classes (col-start-2, row-start-1) make any difference now that you are not using a grid?

petebacondarwin (approved):

frontend/src/components/AccountCard/AccountCard.tsx
dario-piotrowicz: no I should have removed them 😓

PR #329, opened 2023-02-22
add more limits

xtuc:


backend/src/activitypub/actors/index.ts
```diff
@@ -43,32 +43,41 @@ export async function get(url: string | URL): Promise<Actor> {
 
 	const data = await res.json<any>()
 	const actor: Actor = { ...data }
-	actor.id = new URL(data.id)
+	actor.id = new URL(actor.id)
 
-	if (data.content) {
-		actor.content = await sanitizeContent(data.content)
```
xtuc, line 49: @petebacondarwin @dario-piotrowicz  this changes content -> summary. Why were we using content before? bad copy/pasteing from object sanitization?

dario-piotrowicz: yeah it looks like a mistake, I think Pete added it, he can confirm 🙂 

although, is this for accounts?
is `summary` the correct field? shouldn't it be `note`?
Screenshot 2023-02-22 at 14 33 10

xtuc: `note` is the Mastodon term for `summary`

petebacondarwin: Yeah my fault. I misunderstood the model.

PR #338, opened 2023-02-23
Changed to use getErrorHtml

yug1224:
Changed to use `getErrorHtml`.

Could you please check this.

| Before | After |
| :--: | :--: |
| before | after |

The following was used as a reference.

https://github.com/cloudflare/wildebeest/blob/c55b09a7718bb65a7d3410fdfbb9dc7cc369a91a/frontend/src/routes/(frontend)/explore/index.tsx#L17-L20


dario-piotrowicz (approved):
Awesome, thank you very much @yug1224! :smiley:

Issue #343, opened 2023-02-25
Migration step fails but shows as successful run in Action UI


sterlingwes:
Was getting "timeline cannot be loaded, please try again" when I tried to load the site after deploy and couldn't figure out why. Turns out the database was empty! Even though the deploy job seemed to finish fine first time around:

<img width="1103" alt="Screenshot 2023-02-25 at 6 09 14 PM" src="https://user-images.githubusercontent.com/1047502/221383740-9b4959a8-2b77-4a17-8368-3582b48aaab3.png">

<img width="906" alt="Screenshot 2023-02-25 at 6 09 24 PM" src="https://user-images.githubusercontent.com/1047502/221383742-f565ffb3-6790-4f67-a717-e1a1bc96a65a.png">

Looks like the d1 migrations command doesn't exit with a non-zero?

The fix for me was to pull out the migration steps in a separate workflow so I could run it again (https://github.com/cloudflare/wildebeest/compare/main...sterlingwes:wildebeest:fix/migration)

xtuc:
Intermittent D1 error, should fix itself after a rerun (which is what your change did).

sterlingwes:
@xtuc the issue is that the error doesn't manifest as a GH action run failure, which is the actual issue here

sterlingwes:
@xtuc this is still an issue

should i open this on the wrangler repo? unclear to me where this exit code is getting swallowed

PR #344, opened 2023-02-25
chore(docs): add troubleshooting for server errors

sterlingwes:
Found it really hard to get basic info on why I was seeing an error page when I loaded the app. Figure these steps might be useful. Would love any tips on how to get the deployment ID in a nicer way than directing them to guess at UDIDs in the URL bar!

These steps helped me get at the problem caused by https://github.com/cloudflare/wildebeest/issues/343


docs/troubleshooting.md
```diff
@@ -19,6 +19,17 @@ If you attempted to deploy Wildebeest in your account and something failed, or y
...
 
+### Finding error causes
+
+If you get an error page or 500 status error for the main HTML document you can find the source of the error using the wrangler CLI:
+
+- go to your Cloudflare account dashboard > Pages and click on `wildebeest-YOURNAME`
+- click on the Production deployment for your project (View details)
+- copy the second UUID in the URL, which is your deployment ID
+- get wrangler CLI if you don't already have it (`yarn global add wrangler`)
+- run `wrangler pages deployment tail DEPLOYMENT_ID_HERE`
```
dario-piotrowicz, line 30: the deployment id is optional and people can just do `wrangler pages deployment tail`, then select their project via the interactive CLI and it will tail their latest deployment, maybe we should suggest that as it sounds simpler got newcomers?

docs/troubleshooting.md
```diff
@@ -19,6 +19,17 @@ If you attempted to deploy Wildebeest in your account and something failed, or y
...
 You can now start a clean install.
 
+### Finding error causes
+
+If you get an error page or 500 status error for the main HTML document you can find the source of the error using the wrangler CLI:
+
+- go to your Cloudflare account dashboard > Pages and click on `wildebeest-YOURNAME`
+- click on the Production deployment for your project (View details)
+- copy the second UUID in the URL, which is your deployment ID
+- get wrangler CLI if you don't already have it (`yarn global add wrangler`)
```
dario-piotrowicz, line 29: `yarn` isn't the default node package manager, so I would go with `npm install -g wrangler` instead (you can also could include the `yarn` and `pnpm` versions as well if you think that that would be valuable)

dario-piotrowicz:
left a few comments regarding tailing, by the way thanks a bunch @sterlingwes :smiley:

docs/troubleshooting.md
sterlingwes: I got an error if I didn't specify the deployment ID. Did that change recently?

docs/troubleshooting.md
sterlingwes: yarn is required by this project? (`yarn.lock`)

docs/troubleshooting.md
sterlingwes: hmm can't reproduce now, will update 👍🏻

sterlingwes:
Updated! hope that works. thanks for the weekend review 🙏🏻

docs/troubleshooting.md
dario-piotrowicz: ah... mh.... yeah you do have a point....  but people (like me! :grin:) can use yarn locally for the project and just use `npm` globally, anyways I've seen that you've changed it to refer to the cloudflare docs, great idea! :smiley:

dario-piotrowicz:
> Updated! hope that works. thanks for the weekend review 🙏🏻

No problem, thank you very much for contributing! :smiley: 

Ready to merge, just one small last thing, could you squash the commits into one? :pray:

sterlingwes:
You should really consider changing the default to squash on merge so that force pushing on branches during a review isn't necessary

from my other PR:

>have you considered changing the default for the repo to squash on merge? That seems to be the standard nowadays since it encourages PRs to be about one thing (leading to one commit on main, but still allowing for granular commits on the branch and not interrupting the review with force pushes) which is a good practice

>Also lowers the contribution barrier to entry.. i find myself less enthused about contributing anything substantial to this repo when it's so much back and forth over smol readme changes

PR #348, opened 2023-02-28
handle hashtags in statuses

### LINKED ISSUE #345: Still having issues that Hashtags don't work
https://github.com/cloudflare/wildebeest/issues/243 closed by @xtuc 3 days ago, but please see following toot as evidence that this is not fixed:

https://nptcomputing.xyz/@nathan/a76689a7-fb09-407c-9fd9-ddd7f19eb8f4/

I am up to date with commits and look to be running v.98d0e1b.

Hashtags are not underlined in this post.  They are not clickable (a) on the instance, (b) in Mastodon Android App and (c) Pinafore.

Please could we get some evidence of testing before an issue is closed.

dario-piotrowicz:
- by parsing them in enrichStatus
 - implementing a tag page to show their timeline

resolves #345

### DIFF
--- a/backend/test/mastodon.spec.ts
+++ b/backend/test/mastodon.spec.ts
@@ -198,6 +198,51 @@ describe('Mastodon APIs', () => {
 				assert.equal(enrichStatus(`@!@£${link}!!!`, []), `<p>@!@£<a href="${link}">${urlDisplayText}</a>!!!</p>`)
 			})
 		})
+
+		test('convert tags to HTML', async () => {
+			const tagsToTest = [
+				{
+					tag: '#test',
+					expectedTagAnchor: '<a href="/tags/test" class="status-link hashtag">#test</a>',
+				},
+				{
+					tag: '#123_joke_123',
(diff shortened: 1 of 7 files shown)


backend/test/mastodon.spec.ts
```diff
@@ -198,6 +198,52 @@ describe('Mastodon APIs', () => {
...
+					tag: '#_123',
+					expectedTagAnchor: '<a href="/tags/_123" class="status-link hashtag">#_123</a>',
+				},
+				{
+					tag: '#example:',
+					expectedTagAnchor: '<a href="/tags/example" class="status-link hashtag">#example</a>:',
+				},
+				{
+					// IS THIS OK?
+					tag: '#tagA#tagB',
```
dario-piotrowicz, line 222: @xtuc what do you think? is this ok? should we recognize both tags? or should this just fail to recognize any?

xtuc: Tried with Mastodon and they only convert the first one into a link:
image


I think converting both to links is fine too.

xtuc (approved):
fine to merge as is, thanks

PR #360, opened 2023-03-01
Register clients without a website

kelvin27315:
in Mastodon: https://github.com/mastodon/mastodon/blob/main/app/lib/application_extension.rb


xtuc (approved):
Thanks!

PR #361, opened 2023-03-02
Enable about and settings pages + other small tweaks/improvements

dario-piotrowicz:


frontend/src/routes/(frontend)/about/index.tsx
```diff
@@ -27,9 +25,7 @@ type AboutInfo = {
 	}
 }
 
-export const aboutInfoLoader = loader$<Promise<AboutInfo>>(async ({ resolveValue, request, platform, html }) => {
-	throw html(404, getNotFoundHtml())
```
dario-piotrowicz, line 31: here and for the settings page I thought of having an env and based on that 404ing or not

but I figured that it'd be a bit overengineered and not really that needed/useful (we can check social.eng first before synching cloudflare.social)

(also AFAIK we don't even provide a way to access those pages besides directly going to them via their url, so people probably won't try to access them)

frontend/src/routes/(admin)/settings/server-settings/about/index.tsx
```diff
@@ -16,7 +16,7 @@ export const action = action$(async (data, { request, platform }) => {
 	let success = false
 	try {
 		const response = await handleRequestPost(
-			getDatabase(platform),
+			await getDatabase(platform),
```
petebacondarwin, line 19: How did this work in the first place??

dario-piotrowicz: this page wasn't being exposed so I think it just didn't work 😢 

(that's my bad, I rebased a to main and got the promise based getDatabase but must have missed to update it here 😓 )

Issue #269, opened 2023-02-13
Cannot logon with Mastodon mobile client 


kelsallp:
I created my userid with Pinafore and OAuth.

I have just tried to logon with Mastodon mobile App. The app asks for the server name, and then shows a new user logon screen, rather than a logon screen (OAuth).

xtuc:
Which app are you using?

kelsallp:
It is the Mastodon Android mobile app.

I have just tried it with the Pinafore PWA app in Chrome on Android - same error.

kelsallp:
No it does not work. Here is what I get:

1) Mastodon mobile app loads
2) Mastodon app asks for server name. This wildebeest server name is given, and the app allows this to be selected
3) The Wildebeest Login page is shown (from the page title), but the option to login is not shown, only the option to register
4) If I do actually enter my user name (to register again), then the next screen shows 'Internal Server Error'

kelsallp:
Works now - pops up the CF OAuth logon screen, and goes to a logged-on state after selecting the account.

Issue #253, opened 2023-02-10
Incomplete Mastodon API endpoint implementation breaks 3rd-party app OAuth flow


DataDrivenMD:
Elk, along with several other 3rd-party apps, call `GET /api/v1/apps/verify_credentials` when they first register themselves with a Mastodon-compatible server (i.e. the app, not the user). This endpoint is not implemented in Wildebeest. This breaks OAuth flow during initial app registration.

In Elk, the missing endpoint triggers the following error:

```{"url":"/api/wildebeest.example.com/oauth?origin=https://elk.example.com?code=xxxxxx","statusCode":400,"statusMessage":"App not registered for server: wildebeest.example.com","message":"App not registered for server: wildebeest.example.com","stack":""}```

DataDrivenMD:
Adding a clarification that this issue (#253) is distinct from [issue #35](https://github.com/cloudflare/wildebeest/issues/35) -- the incompatibility with IceCubes App described in issue #35 stems from an incomplete `GET /api/v2/instance` implementation. In the current issue, here, the problem stems from a missing `GET /api/v1/apps/verify_credentials` endpoint.

Issue #35, opened 2023-01-07
Some clients not supported


penalosa:
I imagine this is known, but to list some that I've tried which don't work:
- https://mastodon.social/@MonaApp/109506252875971252 (fails to log in)
- https://mastodon.cloud/@icecubesapp (shows an "instance not supported" message)

xtuc:
Have you tried since? We did many changes to support other clients.

riaf:
I was interested in this issue so I did a little bit of research on IceCubesApp. I noticed that the "instance not supported" message is no longer there, but the sign in button was still not active. I think this issue can be resolved by adding the "configuration" attribute to the `/api/v1/instance`, as I was able to proceed further by adding that attribute in [my repository](https://github.com/riaf/wildebeest/commit/1f12131eb6d546235ef4dbee0f51f2c93f0a8405) (although the data was not correct, though haha).

However, I have not thoroughly investigated the internal yet, but it seems that the POST request to `/api/v1/apps` is being sent as a form instead of JSON and the Content-Type is missing from the request header, causing it to not be handled correctly.

So, I think that fixing this issue may require both modifications in the wildebeest project and in each of the client apps.

riaf:
Sorry for forgetting to report this. I just realized that IceCubesApp cannot handle the response from a POST request to `/api/v1/apps` correctly because there is no `id` attribute in the response.

DataDrivenMD:
I can confirm that IceCubes app remains incompatible. The problem seems to stem from an incomplete implementation of the `GET api/v2/instance` endpoint (as previously mentioned by @riaf. The app hangs while trying to retrieve the instance metadata, though it's not entirely clear which attributes it needs to proceed.

In any case, once that roadblock is cleared, I expect that we'll hit this other roadblock I filed under issue #253

bretcarmichael:
Ivory for iOS by Tapbots is also incompatible. After authentication, a message is returned the it (Wildebeest) is not a Mastodon server.

DataDrivenMD:
I reached out to the Tapbots team to ask how they determine whether the server is Mastodon-compatible. Awaiting their response.

DataDrivenMD:
Head back from the Tapbots team. They said they require post/statuses/toots to have numerical IDs. Wildebeest uses UUIDs.

Separately, while digging into the issue, I noticed that Wildebeest uses the ActivityPub URI as the Mastodon account number. Clients that expect an integer value are going to reject this, too.

[...]

PR #363, opened 2023-03-02
fix: `/api/v1/instance`

DataDrivenMD:
Taken together, this PR will improve compatibility with 3rd-party apps, including IceCubes and Ivory

- [X] [Refactor existing test suite for instance-level Mastodon API endpoints](https://github.com/cloudflare/wildebeest/commit/8fb9b9b100cab7ee45d2f60a6786dcf6d700e225): the intent is to align the testing suite with the API schema to improve organization of testing coverage across Mastodon API versions
- [X] [Fix bug in getAdmins() method and align `/api/v1/instance` endpoint with Mastodon API spec](https://github.com/cloudflare/wildebeest/commit/2e303528215c368e1ea6ec2b47808cf9c0f3b606): the bug was causing no admins to be returned under certain conditions; the lack of testing allowed this bug to fly under the radar until the value was called by the `/api/v1/instance` handler
- [X] [Added test to ensure that statistics returned from `/api/v1/instance` are accurate](https://github.com/cloudflare/wildebeest/commit/0f2d29ba5e8b80fdc43b597343bb6740270cb845)

Relevant issues: 
- https://github.com/cloudflare/wildebeest/issues/35
- https://github.com/cloudflare/wildebeest/issues/253
- https://github.com/cloudflare/wildebeest/issues/269


DataDrivenMD:
@dario-piotrowicz Every playwright test from `seo.spec.ts` is failing but it's not clear to me why that's happening since I didn't touch the front-end code in this PR. Separately, I'm getting 4-6 playwright failures that don't seem to have any discernible reason-- when I manually follow the tests, the results are as expected; so, the assertions are failing prematurely. 

Is there another branch that I should be using as the base for this PR?

cc @xtuc

DataDrivenMD:
So, I reverted all the changes in this PR, updated from `cloudflare/wildebeest/main` and then re-ran the playwright tests, and 6 tests *still* failed. Then, I manually checked one of the failing tests, and the page rendered correctly. So, it seems that there's something wrong with the playwright configuration (perhaps the timeout period is too short?). 

cc @xtuc @dario-piotrowicz 

playwright_failing
manual_validation

functions/api/v1/instance.ts
```diff
@@ -1,36 +1,84 @@
...
-	res.thumbnail = DEFAULT_THUMBNAIL
-
-	// Registration is disabled because unsupported by Wildebeest. Users
-	// should go through the login flow and authenticate with Access.
-	// The documentation is incorrect and registrations is a boolean.
-	res.registrations = false
+	const adminActors = await getAdmins(db)
+	if (adminActors.length === 0) {
+		console.error('Server misconfiguration: missing admin account')
+		return error.internalServerError()
```
DataDrivenMD, line 31: This was causing playwright to return without adding SEO tags

DataDrivenMD:
Update: I fixed the problem by [adjusting Playwright timeouts](https://github.com/cloudflare/wildebeest/pull/375)...and that surfaced an uncaught exception in my code, which I fixed.

backend/src/mastodon/sql/instance.ts
```diff
@@ -0,0 +1,10 @@
+// Prepared statements for Mastodon Instance API endpoints
+export const instanceStatisticsQuery = (origin: string): string => {
```
xtuc, line 2:
```suggestion
export const instanceStatisticsQuery = (domain: string): string => {
```

For consistency, we called it `domain`. Could you please rename them?

config/versions.ts
```diff
@@ -5,6 +5,8 @@ export const MASTODON_API_VERSION = '4.0.2'
 
 export const WILDEBEEST_VERSION = packagejson.version
 
-export function getVersion(): string {
-	return `${MASTODON_API_VERSION} (compatible; Wildebeest ${WILDEBEEST_VERSION})`
+export function getVersion(domain?: string): string {
+	return `Wildebeest/${WILDEBEEST_VERSION} (Mastodon/${MASTODON_API_VERSION} compatible; +https://${
+		domain ?? 'github.com/cloudflare/wildebeest'
+	})`
```
xtuc, line 11: Why is this change needed?  This looks like the User-Agent function:
```
export function getFederationUA(domain: string): string {
	return `Wildebeest/${WILDEBEEST_VERSION} (Mastodon/${MASTODON_API_VERSION}; +${domain})`
}
```

functions/api/v1/instance.ts
```diff
@@ -1,36 +1,93 @@
...
+	const res: MastodonInstance = {
+		uri: domain,
+		title: env?.INSTANCE_TITLE,
+		description: env?.INSTANCE_DESCR,
+		short_description: env?.INSTANCE_DESCR,
+		email: env?.ADMIN_EMAIL,
+		version: getVersion(domain),
+		languages: ['en'],
+		registrations: env?.INSTANCE_ACCEPTING_REGISTRATIONS ?? false,
+		approval_required: env?.INSTANCE_REGISTRATIONS_REQUIRE_APPROVAL ?? false,
```
xtuc, line 37: registrations / approval_required should be hardcoded to false for now

backend/src/types/env.ts
```diff
@@ -21,6 +21,9 @@ export interface Env {
 	INSTANCE_DESCR: string
 	VAPID_JWK: string
 	DOMAIN: string
+	INSTANCE_ACCEPTING_REGISTRATIONS?: boolean
+	INSTANCE_REGISTRATIONS_REQUIRE_APPROVAL?: boolean
+	INSTANCE_CONFIG_STATUSES_MAX_CHARACTERS?: number
```
xtuc, line 26: Not sure INSTANCE_CONFIG_STATUSES_MAX_CHARACTERS is useful because we don't respect it yet. It's hardcoded to 500 (or some value) at the moment.

functions/api/v1/instance.ts
```diff
@@ -1,36 +1,93 @@
...
+		invites_enabled: false,
+		thumbnail: DEFAULT_THUMBNAIL,
+		configuration: {
+			statuses: {
+				max_characters: env?.INSTANCE_CONFIG_STATUSES_MAX_CHARACTERS ?? 500,
+				max_media_attachments: 4,
+				characters_reserved_per_url: 23,
+			},
+			media_attachments: {
+				supported_mime_types: ['image/jpeg', 'image/png', 'image/gif', 'image/webp', 'video/mp4'],
```
xtuc, line 47: We don't support uploading videos at the moment.

functions/api/v1/instance.ts
```diff
@@ -1,36 +1,93 @@
...
+				image_size_limit: 10485760,
+				image_matrix_limit: 16777216,
+				video_size_limit: 41943040,
+				video_frame_rate_limit: 60,
+				video_matrix_limit: 2304000,
+			},
+			polls: {
```
xtuc, line 54: We don't support polls at the moment

dario-piotrowicz:
@DataDrivenMD I'm so sorry I didn't manage to reply earlier 🙇 

___

Regarding SEO, it should work on main, is your branch up to date with main? and/or have you solved the issue?

___

Regarding increasing the timeouts I really am not a big fan of that because the current timeouts were working just fine so if you need to increase them it would suggest that our code is getting slower or something like that (also they were longer at some point but it was getting annoying that we'd have to wait a while to see if they would fail so I did shorten the timeouts)

`and that surfaced an uncaught exception in my code, which I fixed.` 👈  I am not really sure how longer timeouts would surfaced the issue, can you tell me? 🙂 
(and regarding this, before increasing the timeouts, have you tried viewing the remote playwright reports? didn't that help?)

backend/src/mastodon/sql/instance.ts
DataDrivenMD: Absolutely. I also added inline documentation to clarify that the parameter expected is an **origin** (i.e. must include the `https://`)

config/versions.ts
DataDrivenMD: You're right, it doesn't seem necessary. I *think* (but can't be certain) that when I wrote this code, this file contained the UA method, and the file that is now `./ua.ts` was named `./constants.`. In any case, this proposed change is meant to serve the same purpose as the `getFederationUA()` method, so I'll migrate my proposed changes to that file instead. Here's the rationale for the proposed change:

It's intended as a courtesy to Mastodon instance admins that expect the UA string for ActivityPub hosts to be structured in a specific way.* Speaking from my own experience, I use a combination of HTTP request headers that include the UA string as a custom version of Cloudflare's "Browser Integrity Check".

Here are some real-world examples:
`http.rb/5.1.1 (Mastodon/4.1.0; +https://mastodon.social/)`
`http.rb/5.1.0 (Mastodon/4.0.2; +https://mastodon.sdf.org/)`
`http.rb/5.1.1 (Mastodon/4.1.0+glitch; +https://ramblings.social/)`
`AodeRelay (ap-relay/v0.3.79-HEAD-4998cd3a; +https://relay.universeodon.com)`
`Akkoma 3.6.0-0-g45a11aa-develop; https://ti22.pro <admin@ti22.pro>`
`Pleroma 2.5.50-190-g4c0ab884.develop+soapbox; https://social.nettohikari.com <admin@nettohikari.com>`

backend/src/types/env.ts
DataDrivenMD: Agreed. When I wrote this code, the work on the Admin component hadn't been merged to `main` so all of these `ENV` vars I added are a moot point. I'll remove all of them for now. `INSTANCE_CONFIG_STATUSES_MAX_CHARACTERS` may find its way back in [PR #366](https://github.com/cloudflare/wildebeest/pull/366) because that PR deals with several validation methods for ActivityPub and Mastodon objects

functions/api/v1/instance.ts
DataDrivenMD: Removed `'video/mp4'`

functions/api/v1/instance.ts
DataDrivenMD: Some Mastodon clients will throw indecipherable errors if they don't receive a valid object for the `polls` property. I changed the parameters to get beyond typechecking validation steps in such clients, while also making it clear that polls are not yet supported.

functions/api/v1/instance.ts
DataDrivenMD: Done!

functions/api/v1/instance.ts
```diff
@@ -1,36 +1,93 @@
...
 export const onRequest: PagesFunction<Env, any> = async ({ env, request }) => {
-	const domain = new URL(request.url).hostname
-	return handleRequest(domain, env)
+	const domain: string = new URL(request.url).hostname
+	const dbOverride: Database = await getDatabase(env)
+	return handleRequest(domain, env, dbOverride)
 }
 
-export async function handleRequest(domain: string, env: Env) {
+export async function handleRequest(domain: string, env: Env, dbOverride?: Database) {
```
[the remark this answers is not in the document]
DataDrivenMD: The reason this is optional, here, is due to a quirk in the playwright tests: the `seo.spec.ts` calls the `/api/v1/instance` endpoint without pulling the `Database` instance binding into the tests' context, so those 20 tests will fail. Meanwhile, the unit tests do not have access to the full `EventContext` which is available to the `PagesFunction` at runtime when deployed to Cloudflare.

Making the `Database` parameter an optional *override* was my workaround for this conundrum. A few lines down, I check to see if the override exists, if it's not, then I `try` to extract the `Database` binding from the `env` parameter (`await getDatabase(env)`). If that throws an error (as playwright e2e testing does), it is caught and treated as if a `Database` binding isn't available.

As an aside, the `Database` binding is necessary to run the query that returns instance-level statistics for `user_count`, `status_count`, and `domain_count`. These specific parameters are necessary for IceCubes app compatibility.

DataDrivenMD:
> @DataDrivenMD I'm so sorry I didn't manage to reply earlier 🙇

No worries! Thanks for closing the loop.

> `and that surfaced an uncaught exception in my code, which I fixed.` 👈 I am not really sure how longer timeouts would surfaced the issue, can you tell me? 🙂 (and regarding this, before increasing the timeouts, have you tried viewing the remote playwright reports? didn't that help?)

Prior to extending the timeouts I would see different e2e tests fail with every run (i.e. every time I called `yarn playwright test`). Without changing a single line of code, I ran the tests several times: sometimes 50+ tests would fail, sometimes 30-40, then 26, then 56. It was all over the place, and the tests in `seo.spec.ts` wouldn't always fail as a group.

Once I extended the timeout period, the playwright tests defined in other files (i.e. not `seo.spec.ts` were able to run completion (and `PASS`). Then, and only then, was I able to get consistent results between runs: all of the tests in `seo.spec.ts` began failing as group in a repeatable manner. That's what I mean by "surfaced an uncaught exception in my code".

IOW: the e2e tests were working as designed but the short timeout period triggered some of them to fail prematurely in a flaky manner. This, in turn, made it impossible to figure out whether the tests were failing due a bug in my code vs. failing due to resource constraints vs. some other reason. Once the timeout period was extended, it was possible to see that a) there was an actual bug in the code and b) trace the exception. The timeout issue is fixed in [PR #375](https://github.com/cloudflare/wildebeest/pull/375), which has now been merged to `main`

PS - Sorry for the confusion, and I hope my answer clarifies the issue

DataDrivenMD:
@xtuc Just want to flag that this PR is ready for re-review. I left two comments unresolved but added explanations for the proposed changes.

functions/api/v2/instance.ts
```diff
@@ -19,7 +19,7 @@ export async function handleRequest(domain: string, db: Database, env: Env) {
 	const res: InstanceConfigV2 = {
 		domain,
 		title: env.INSTANCE_TITLE,
-		version: getVersion(),
+		version: getFederationUA(domain),
```
xtuc, line 22: Is this change fixing something? That would be a breaking change and I don't see a reason to do it

functions/api/v1/instance.ts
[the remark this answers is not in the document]
dario-piotrowicz: sorry I am confused, the playwright tests do rely on your local D1Database (otherwise you wouldn't see any toots, accounts, etc...) so I am not sure I'm understanding the issue 😕 

also playwright is totally unrelated to the application, it practically runs on a separate process and tests your local server, so I am not sure how playwright can affect the app's behaviour in any ways 😕

functions/api/v2/instance.ts
DataDrivenMD: I think I misunderstood what you were asking for in your last review when you asked why I was suggesting changes to the `getVersion()` method instead of using the `getFederationUA(domain: string)` method. I just reverted this change, and the related change to `wildebeest/config/versions.ts`. I *think* that will yield what you're for, but if I still didn't get it right can you please suggest specific changes or offer an example to clarify the question?

functions/api/v1/instance.ts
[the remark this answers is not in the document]
DataDrivenMD: @dario-piotrowicz What you're saying makes sense, and aligns with what I'm seeing. Here's another way to phrase the question: does playwright make the DB binding available via the mocked `EventContext`? If so, then I should be able to run the following code without throwing an exception:
```js
export async function handleRequest(domain: string, env: Env) {
  const db: Database = await getDatabase(env)
}
```

The problem is that running that code throws the following error:
```
GET /explore 200 OK (10.28ms)
TypeError: Invalid state: WritableStream is locked
    at new NodeError (node:internal/errors:371:5)
    at WritableStream.close (node:internal/webstreams/writablestream:210:9)
    at /<REDACTED>/wildebeest/frontend/server/entry.cloudflare-pages.js:20:2366
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at Object.S2 [as next] (/<REDACTED>/wildebeest/frontend/server/entry.cloudflare-pages.js:20:4433)
    at ze3 (/<REDACTED>/wildebeest/frontend/server/entry.cloudflare-pages.js:20:6547) {
  code: 'ERR_INVALID_STATE'
}
```

DataDrivenMD:
Turns out there were 2 issues causing e2e tests to fail (now fixed):
1. I needed to pass the `DOMAIN=cloudflare.com` binding to the e2e test suite in order to pull the instances stats. I added that binding to the `yarn ci-dev-test-ui` script in the `package.json` file.

2. I needed to update the type definition used by the frontend to match the output from the Mastodon-compatible API

Separately, the TypeError that I shared above is still there but all tests are passing, so I don't know what to make of it. It seems to stem from the query builder code.

cc @dario-piotrowicz @xtuc

DataDrivenMD:
Just upstreamed all the changes in this branch. Should be ready again for you @xtuc

functions/api/v1/instance.ts
```diff
@@ -1,36 +1,99 @@
...
-export async function handleRequest(domain: string, env: Env) {
+export async function handleRequest(domain: string, env: Env, db: Database) {
 	const headers = {
 		...cors(),
 		'content-type': 'application/json; charset=utf-8',
 	}
 
-	const res: any = {}
+	if (env.ADMIN_EMAIL === 'george@test.email') {
+		db = await getDatabase(env)
```
xtuc, line 30: why do you need this? db is already passed in the function, do you still need to call getDatabase?

functions/api/v1/instance.ts
```diff
@@ -1,36 +1,99 @@
...
+		const message: string = `Invalid request: 'domain' (${domain}) !== 'env.DOMAIN' (${env.DOMAIN})`
+		console.trace(message)
+		return error.validationError(message)
+	}
 
-	// Registration is disabled because unsupported by Wildebeest. Users
-	// should go through the login flow and authenticate with Access.
-	// The documentation is incorrect and registrations is a boolean.
-	res.registrations = false
+	const statsDomain: string = env.ADMIN_EMAIL === 'george@test.email' ? '0.0.0.0' : domain
```
xtuc, line 43: I prefer to avoid test specific behaviors in prod endpoints

PR #365, opened 2023-03-03
fix: non-compliant Mastodon Account endpoints

DataDrivenMD:
## Completed
- [X] [Define CustomEmoji type](https://github.com/cloudflare/wildebeest/commit/8d9daec481d781f9799fe2e9d736d9c8e08ada96): formally defining this type helps with serialization of a Mastodon API-compliant version of the [Mastodon `Account`](https://docs.joinmastodon.org/entities/Account/) entity
- [X] [Add Mastodon ID utilities](https://github.com/cloudflare/wildebeest/commit/d4ad37573404c471a8a9da59d857c5d6112a77ae): these methods are necessary to generate numeric IDs that resemble those that are used in the [official Mastodon project](https://github.com/mastodon/mastodon/blob/main/lib/mastodon/snowflake.rb). Several of the most popular 3rd-party Mastodon apps, including Ivory, rely on these numeric account IDs. So, these methods lay the foundation for greatly-improved compatibility with 3rd-party apps.
- [X] [MastodonAccount *mostly* aligns w/ Masto API spec](https://github.com/cloudflare/wildebeest/commit/3594448fb333bb11e558d578bfb72556e6a0c3e1) and [Add /api/v1/accounts/lookup + tests](https://github.com/cloudflare/wildebeest/commit/db4a6b4baf1ef1c11a54962f78be6d35efdf3685): these two commits bring the changes in this PR together. One key change here is the addition of the `/api/v1/accounts/lookup` endpoint, which was previously missing, and was contributing to various 3rd-party app incompatibilities. With these changes, I was able to refactor the accounts test suite, which was necessary because 3 tests were miscategorized as belonging to `/api/v1/accounts/:id` when they should have actually been tests for `/api/v1/accounts/lookup`. 

## WIP

- [ ] Generating a Mastodon account ID when an AP actor is added to the `actors` table
- [ ] DB migration to ensure that every entry in the `actors` table has a mastodon id
- [ ] Update the code in `wildebeest/functions/api/v1/accounts/[id].ts` and `wildebeest/functions/api/v1/accounts/[id]/*.ts` so that they actually use the Mastodon Account ID instead of the ActivityPub actor ID because as things stand now, the implementation is not-compliant with the Mastodon API spec, and it's breaking compatibility with several popular 3rd-party apps
- [ ] Update test suite `/api/v1/accounts/:id` endpoints

cc @xtuc

### DIFF
--- a/backend/src/accounts/getAccount.ts
+++ b/backend/src/accounts/getAccount.ts
@@ -8,6 +8,7 @@ import { queryAcct } from 'wildebeest/backend/src/webfinger/index'
 import { loadExternalMastodonAccount, loadLocalMastodonAccount } from 'wildebeest/backend/src/mastodon/account'
 import { MastodonAccount } from '../types'
 import { adjustLocalHostDomain } from '../utils/adjustLocalHostDomain'
+import { findMastodonAccountIDByEmailQuery } from 'wildebeest/backend/src/mastodon/sql/account'
 
 export async function getAccount(domain: string, accountId: string, db: Database): Promise<MastodonAccount | null> {
 	const handle = parseHandle(accountId)
@@ -45,4 +46,10 @@ async function getLocalAccount(domain: string, db: Database, handle: Handle): Pr
 	}
 
 	return await loadLocalMastodonAccount(db, actor)
+}
+
+export async function getAccountByEmail(domain: string, email: string, db: Database): Promise<MastodonAccount | null> {
+	const row: any = await db.prepare(findMastodonAccountIDByEmailQuery).bind(email).first()
+
+	return await getAccount(domain, row?.id, db)
 }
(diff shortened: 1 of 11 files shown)


backend/src/mastodon/status.ts
```diff
@@ -165,12 +175,15 @@ export async function toMastodonStatusFromRow(domain: string, db: Database, row:
...
+		const mastodonAccount: MastodonAccount | null = await loadExternalMastodonAccount(acct, author)
+		if (mastodonAccount === null) {
+			return null
+		}
 
 		// Restore reblogged status
 		status.reblog = {
 			...status,
-			account,
+			account: mastodonAccount,
```
DataDrivenMD, line 186: Note: rest property must be the last in the pattern, and must not have a trailing comma

ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#rest_property

PR #369, opened 2023-03-06
add missing auth loader to settings page and add access check to isUserAdmin

dario-piotrowicz:


frontend/src/routes/(admin)/settings/aliases/index.tsx
```diff
@@ -3,7 +3,8 @@ import { loader$ } from '@builder.io/qwik-city'
 import { checkAuth } from '~/utils/checkAuth'
 
 export const loader = loader$(async ({ request, platform, redirect }) => {
-	const isAuthorized = await checkAuth(request, platform)
+	const jwt = request.headers.get('Cf-Access-Jwt-Assertion') || ''
```
xtuc, line 6: this should be using CF_Authorization in cookies as the route isn't running Access

frontend/src/routes/layout.tsx
```diff
@@ -9,7 +9,8 @@ type AccessLoaderData = {
 }
 
 export const accessLoader = loader$<Promise<AccessLoaderData>>(async ({ platform, request }) => {
-	const isAuthorized = await checkAuth(request, platform)
+	const jwt = request.headers.get('Cf-Access-Jwt-Assertion') || ''
```
xtuc, line 12: same; should be CF_Authorization in cookies

frontend/src/routes/(admin)/settings/aliases/index.tsx
dario-piotrowicz: ah ok thanks I thought so too but the old implementation was using the header, so I figured that maybe I was wrong 😓 👍

PR #370, opened 2023-03-06
add type-check frontend script

dario-piotrowicz:
add such prelint script so that it runs tsc making sure that all typings are valid

### DIFF
--- a/backend/src/database/neon.ts
+++ b/backend/src/database/neon.ts
@@ -4,7 +4,7 @@ import type { Env } from 'wildebeest/backend/src/types/env'
 
 function sqliteToPsql(query: string): string {
 	let c = 0
-	return query.replaceAll(/\?([0-9])?/g, (match: string, p1: string) => {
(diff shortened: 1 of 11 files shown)


backend/src/database/neon.ts
```diff
@@ -4,7 +4,7 @@ import type { Env } from 'wildebeest/backend/src/types/env'
 
 function sqliteToPsql(query: string): string {
 	let c = 0
-	return query.replaceAll(/\?([0-9])?/g, (match: string, p1: string) => {
+	return query.replace(/\?([0-9])?/g, (match: string, p1: string) => {
```
xtuc, line 7: why this change?

dario-piotrowicz: it produces a typings error:
Screenshot 2023-03-06 at 15 44 47


But anyways if the regex ends with `/g` then it applies the replacement to all the instances so I think this is ok, do you think it's not?

xtuc: Yeah make sense.

PR #375, opened 2023-03-07
fix: e2e tests fail due to premature timeouts

DataDrivenMD:
- [X] Fixes timeouts so that tests don't fail on Github Actions runners
- [X] Lowers number of retries so that e2e tests don't run for 2+ hours before giving up

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -19,14 +19,14 @@ const config: PlaywrightTestConfig = {
 		 * Maximum time expect() should wait for the condition to be met.
 		 * For example in `await expect(locator).toHaveText();`
 		 */
-		timeout: process.env.CI ? 5000 : 500,
+		timeout: (process.env.CI ? 30 : 5) * 1000,
 	},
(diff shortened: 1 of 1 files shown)


xtuc (approved):
For the record, I'm not fan of different behaviors between CI and local, but let's merge for now.

dario-piotrowicz:
> For the record, I'm not fan of different behaviors between CI and local, but let's merge for now.
to be fair I added the different behaviour a while back because it was getting annoyed waiting unnecessarily long locally to see a test fail 😅

DataDrivenMD:
> > For the record, I'm not fan of different behaviors between CI and local, but let's merge for now.
> > to be fair I added the different behaviour a while back because it was getting annoyed waiting unnecessarily long locally to see a test fail 😅

I agree, and thank you for making the exception. It was driving me nuts that code would seem to work locally, and then randomly fail during PR checks. I tried a few iterations that relied on just configuration for all environments, but local e2e testing with a single worker is painful to watch lol

PR #395, opened 2023-03-12
Fix status enrichment that includes URLs with single-letter second-level domains

### LINKED ISSUE #394: URLs with single-letter second level domain are not enriched
URLs with single-letter second level domain (like https://g.co, https://t.co) are not enriched. The length of any one label in domain is limited to between 1 and 63 octets in [RFC 2181](https://datatracker.ietf.org/doc/html/rfc2181).

[Image from Gyazo](https://gyazo.com/299a9c2b1ca5728c638e5ca76a92d4fe)

chroju:
This PR fixes #394. Thanks for making great software!

### DIFF
--- a/backend/src/mastodon/microformats.ts
+++ b/backend/src/mastodon/microformats.ts
@@ -11,7 +11,7 @@ function tag(name: string, content: string, attrs: Record<string, string> = {}):
 	return `<${name}${htmlAttrs}>${content}</${name}>`
 }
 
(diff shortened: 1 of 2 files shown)


dario-piotrowicz:
Thanks a bunch for the change @chroju :slightly_smiling_face: 
(and thanks a lot for remembering to add the new cases to the tests too :smiley:)

dario-piotrowicz (approved):

